Issue #247, opened 2014-11-01
Add a cabal build hook to compile the manpage


koalaman:
The Makefile did this, but the Cabal file does not

PR #295, opened 2015-01-24
Add a pre-sdist hook to compile the man page (Github issue #247).

orlitzky:
This replaces the default preSDist hook in Setup.hs with our own. The
only thing the new hook does is compile the man page using callCommand
from System.Process.

If Pandoc fails, the entire sdist process will fail, since
Extra-Source-Files in the cabal file now lists the man page.

This is preferable to a build hook, because Pandoc pulls in a huge
number of dependencies. It's better to build the man page once and
ship it than to require every user to build and install pandoc before
he can build ShellCheck.

This creates another TODO item: an install hook can now be used to
install the man page along with the rest of ShellCheck. But beware,
the "man path" can vary from system to system.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,2 +1,43 @@
-import Distribution.Simple
-main = defaultMain
+import Distribution.PackageDescription (
+  HookedBuildInfo,
+  emptyHookedBuildInfo )
+import Distribution.Simple (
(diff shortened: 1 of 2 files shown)


koalaman:
Awesome, thanks!

PR #361, opened 2015-04-26
Add a JUnit output format style.

deryni:
This code is almost entirely based on the checkstyle format code.

The following links were used in putting the output format together:

    https://confluence.atlassian.com/display/BAMBOO/JUnit+parsing+in+Bamboo
    http://stackoverflow.com/questions/4922867/junit-xml-format-specification-that-hudson-supports
    http://windyroad.com.au/dl/Open%20Source/JUnit.xsd

I should probably also mention that I don't know Haskell at all. =)
This just seemed like a straight-forward enough thing to try to do from examples.

### DIFF
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -89,6 +89,7 @@ formats = Map.fromList [
     ("json", forJson),
     ("gcc", forGcc),
     ("checkstyle", forCheckstyle),
+    ("junit", forJunit),
     ("tty", forTty)
     ]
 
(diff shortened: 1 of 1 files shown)


koalaman:
As discussed, this rather niche use case may be better handled through xslt. I really appreciate the contribution though!

koalaman:
XSLT for transforming Checkstyle to JUnit XML has been [added to the wiki](https://github.com/koalaman/shellcheck/wiki/JUnit)

PR #426, opened 2015-08-06
Cli for shellcheck dot net

icy:
I originally wrote this simple script to use in the `pacapt` project (https://github.com/icy/pacapt/issues/54#issuecomment-128344059). Some contributors may not have `Haskell` on their system, so a web tool is a very quick way to enjoy the tool ;)

Feel free to let me know to make this better.

### DIFF
--- a/utils/webcheck.sh
+++ b/utils/webcheck.sh
@@ -72,7 +72,6 @@ _check_file() {
   }
 
   _simple_check "$_file" || return
-  _perl_check || return
   _shellcheck < "$_file"
 }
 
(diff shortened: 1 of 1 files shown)


koalaman:
Very useful! 

I'm more than happy to let people hit the web service at their own risk, but I'm very hesitant to officially encourage it because shellcheck.net is a simple online demo and not SaaS. I don't want to imply any intentions of keeping it available, responsive or compatible. 

(Also note that while large footprint Haskell tools are required to compile shellcheck, a precompiled binary as installed by a binary package manager does not depend on anything Haskell related)

icy:
Thanks for your update. I understand this completely ;)

Yesterday I tried to run `shellcheck` for all scripts on my `/usr/bin/*` and some personal scripts. The result is here http://icy.theslinux.org/s/shellcheck_results_for_all_scripts_under_usr_bin.log.gz . Hope it's useful for someone ;)

I personally don't want to publish this kind of script, because a mistake can make private data public (People may put their secrets in the script)

PR #553, opened 2015-12-04
Add --color switch

haguenau:
Add --color switch

This changeset adds a --color switch to `shellcheck`. It works in a
similar fashion to the option of the same name in GNU tools such as
`ls`:

* `--color` or, equivalently, `--color=always` prints text
  using color codes, even if stdout is not a terminal;

* `--color=never` disables color;

* `--color=auto`, the same as not specifying a `--color` option at
  all, will print in color when stdout is deemed to be a capable
  terminal.

This switch is useful when using, for example, a color-aware pager:
one can use `shellcheck --color | less -R` to get scrollable, colored
text, whereas `shellcheck` would otherwise print without color codes.

Comments on the code are very welcome; I don't know how well this fits
in the existing codebase, etc. Also I didn't add any tests, although
it certainly would make sense to do so.

# Requesting a pull to haguenau:master from haguenau:add-color-switch
#
# Write a message for this pull request. The first block
# of text is the title and the rest is description.
#
# Changes:
#
# 0044c3d (David Haguenauer, 16 minutes ago)
#    Make use of --color with no argument equivalent to --color=always
#
# a3d4101 (David Haguenauer, 24 hours ago)
#    Add initial support for --color
#
# bd359c5 (David Haguenauer, 24 hours ago)
#    Delete trailing whitespace

### DIFF
--- a/ShellCheck/Formatter/TTY.hs
+++ b/ShellCheck/Formatter/TTY.hs
@@ -35,7 +35,7 @@ format = return Formatter {
     onResult = outputResult
 }
 
-colorForLevel level = 
+colorForLevel level =
     case level of
         "error"   -> 31 -- red
         "warning" -> 33 -- yellow
@@ -44,7 +44,7 @@ colorForLevel level =
         "message" -> 1 -- bold
         "source"  -> 0 -- none
         otherwise -> 0 -- none
-    
+
 outputError file error = do
     color <- getColorFunc
     hPutStrLn stderr $ color "error" $ file ++ ": " ++ error
--- a/ShellCheck/Formatter/TTY.hs
+++ b/ShellCheck/Formatter/TTY.hs
@@ -46,11 +46,11 @@ colorForLevel level =
         otherwise -> 0 -- none
 
 outputError file error = do
-    color <- getColorFunc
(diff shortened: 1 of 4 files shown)


koalaman:
Thanks!

ShellCheck/Formatter/TTY.hs
```diff
@@ -44,13 +44,13 @@ colorForLevel level =
         "message" -> 1 -- bold
         "source"  -> 0 -- none
         otherwise -> 0 -- none
-    
+
 outputError file error = do
-    color <- getColorFunc
-    hPutStrLn stderr $ color "error" $ file ++ ": " ++ error
+    color <- getColorFunc $ ColorAuto -- FIXME: should respect --color
+    hPutStrLn stderr $ color "error ZZZ" $ file ++ ": " ++ error
```
haguenau: Whoops, I accidentally left this in. I'll fix it as soon as I get access to a proper computer.

koalaman: Oh, oops! Don't worry about it, I fixed it in b439041 by adding some formatting options instead of threading it through the result. 

Thanks for the fix!

haguenau: Whew. Thanks to you for being so reactive!

PR #572, opened 2016-01-08
Add a warning when you mkdir with both -p and -m.

eatnumber1:
When using -p, parent directories will not be created with the mode
specified with -m and will instead be created using the default behavior
controlled by umask.

### DIFF
--- a/ShellCheck/Analytics.hs
+++ b/ShellCheck/Analytics.hs
@@ -218,8 +218,8 @@ nodeChecks = [
     ,checkLoopVariableReassignment
     ,checkTrailingBracket
     ,checkNonportableSignals
+    ,checkMkdirDashPM
     ]
-
 
 filterByAnnotation token =
     filter (not . shouldIgnore)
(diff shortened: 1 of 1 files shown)


ShellCheck/Analytics.hs
```diff
@@ -3670,6 +3670,33 @@ checkNonportableSignals _ = checkUnqualifiedCommand "trap" (const f)
...
+prop_checkMkdir1 = verify checkMkdir "mkdir -pm 0755 dir"
+prop_checkMkdir2 = verify checkMkdir "mkdir -vpm 0755 dir"
+prop_checkMkdir3 = verify checkMkdir "mkdir -pm 0755 -v dir"
+prop_checkMkdir4 = verifyNot checkMkdir "mkdir -p dir"
+prop_checkMkdir5 = verifyNot checkMkdir "mkdir -m 0755 dir"
+prop_checkMkdir6 = verifyNot checkMkdir "mkdir dir"
+checkMkdir _ = checkCommand "mkdir" (const $ f (Nothing,Nothing)) where
+    f _ [] = return ()
+    f ((Just mIdent),(Just pIdent)) _ = do warn (getId mIdent) 2174 $ "-m will be ignored for parent directories"
+    f (dashMIdent,dashPIdent) (a:r) =
```
koalaman: You can simplify this with https://github.com/koalaman/shellcheck/blob/master/ShellCheck/ASTLib.hs#L110 

E.g. something like:

    potentially $ do
      dashP <- find ((== "p") . snd) flags
      dashM <- find ((== "m") . snd) flags
      return $ warn (getId $ fst dashM) 2174 "-m will be ignored"
    where flags = getAllFlags cmd in

ShellCheck/Analytics.hs
```diff
@@ -3670,6 +3670,33 @@ checkNonportableSignals _ = checkUnqualifiedCommand "trap" (const f)
...
             "SIGKILL/SIGSTOP can not be trapped."
 
+prop_checkMkdir0 = verify checkMkdir "mkdir -p -m 0755 dir"
+prop_checkMkdir1 = verify checkMkdir "mkdir -pm 0755 dir"
+prop_checkMkdir2 = verify checkMkdir "mkdir -vpm 0755 dir"
+prop_checkMkdir3 = verify checkMkdir "mkdir -pm 0755 -v dir"
+prop_checkMkdir4 = verifyNot checkMkdir "mkdir -p dir"
+prop_checkMkdir5 = verifyNot checkMkdir "mkdir -m 0755 dir"
+prop_checkMkdir6 = verifyNot checkMkdir "mkdir dir"
+checkMkdir _ = checkCommand "mkdir" (const $ f (Nothing,Nothing)) where
```
koalaman: Consider a more specific name like "checkMkdirDashMP" or "checkMkdirIgnoredDashM".

eatnumber1: My thought was that it would work like checkTr, growing to contain multiple checks over time. I'll change it to be more specific though since that seems to be what you prefer.

ShellCheck/Analytics.hs
eatnumber1: I knew there'd be an easier way to do it. I added some logic for the command name. Can you PTAL?

koalaman:
Thanks!

PR #608, opened 2016-02-16
Fix broken links in README

cs-shadow:
Presently, the links pointed to by "GCC error compatibility" and
"CheckStyle compatible XML output" are broken due to improper
way of referring files. This change fixes those links.

Also, this commit removes some trailing whitespaces in README.


koalaman:
Thanks!

PR #616, opened 2016-02-28
add compilation documentation for test runners

rremer:
Super minor documentation update, but for first-time cabal users we can avoid needing to compile twice because 'cabal test' won't run the first time we try.


koalaman:
`cabal test` will still run, just automatically reconfigure right?

koalaman:
I've seen it do this, but now I'm also seeing it not to that. Finally merging, thanks! :P

PR #622, opened 2016-03-02
Add stack support

Arguggi:
I didn't change the README since I wasn't sure if you also wanted to switch that to stack too.


koalaman:
Thanks! Sorry for the slow response.

PR #628, opened 2016-03-11
Fix a handful of typos.

benmwebb:
I noticed a few typos while reading through your README - they should be fixed here.


koalaman:
Thanks!

PR #677, opened 2016-05-26
Add Gentoo to supported distribution list.

Maffblaster:
It was missing from the list. I thought it was best to add it. :)

@koalaman, thanks for all your excellent work! I've found this tool to be a frequent aid in script development!


koalaman:
Thanks!

PR #695, opened 2016-06-18
Properly set the end column for SC2086 findings

eatnumber1:
This change makes the architectural changes needed to support emitting
end columns in AST based analyzers. As a first check, it emits a correct
end column in SC2086.

### DIFF
--- a/ShellCheck/Parser.hs
+++ b/ShellCheck/Parser.hs
@@ -133,6 +133,38 @@ almostSpace =
         char c
         return ' '
 
+withNextId :: Monad m => ParsecT s UserState (SCBase m) (Id -> b) -> ParsecT s UserState (SCBase m) b
+withNextId p = do
+    start <- getPosition
+    id <- createId
+    setStartPos id start
+    fn <- p
+    let t = fn id
+    end <- getPosition
+    setEndPos id end
+    return t
+  where
+    createId = do
+        state <- getState
+        let id = incId (lastId state)
+        putState $ state {
+            lastId = id
+        }
+        return id
+      where incId (Id n) = Id $ n+1
+    setStartPos id sourcepos = do
+      state <- getState
+      let newMap = Map.insert id sourcepos (positionMap state)
+      putState $ state {
+          positionMap = newMap
+      }
+    setEndPos id sourcepos = do
+      state <- getState
+      let newMap = Map.insert id sourcepos (positionEndMap state)
+      putState $ state {
+          positionEndMap = newMap
+      }
+
 --------- Message/position annotation on top of user state
 data Note = Note Id Severity Code String deriving (Show, Eq)
 data ParseNote = ParseNote SourcePos SourcePos Severity Code String deriving (Show, Eq)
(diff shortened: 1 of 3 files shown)


koalaman:
Manually having to mark each end position seems error prone. How about something like 

```
readDoubleQuoted = called "double quoted string" $
  withNextId $ do
    startPos <- getPosition
    doubleQuote
    x <- many doubleQuotedPart
    endPos <- getPosition
    doubleQuote <|> fail "Expected end of double quoted string"
    optional $ do
        try . lookAhead $ suspectCharAfterQuotes <|> oneOf "$\""
        when (any hasLineFeed x && not (startsWithLineFeed x)) $
            suggestForgotClosingQuote startPos endPos "double quoted string"
    return $ \id -> T_DoubleQuoted id x
```

You would then not have to remember to close the scope

eatnumber1:
That's a good idea. Please take another look. I've changed it to reflect your suggestion.

eatnumber1:
ping

ShellCheck/Parser.hs
```diff
@@ -133,6 +133,38 @@ almostSpace =
         char c
         return ' '
 
+withNextId :: Monad m => ParsecT s UserState (SCBase m) (Id -> b) -> ParsecT s UserState (SCBase m) b
```
koalaman: How about `:: Monad m => SCParser m (Id -> b) -> SCParser m b`

ShellCheck/Parser.hs
```diff
@@ -133,6 +133,38 @@ almostSpace =
...
+withNextId p = do
+    start <- getPosition
+    id <- createId
+    setStartPos id start
+    fn <- p
+    let t = fn id
+    end <- getPosition
+    setEndPos id end
```
koalaman: Many of the parsers will also parse and discard trailing spaces, and the end position would then be after all the whitespace. Is there a way around that?

koalaman:
Sorry about the awful latency :(

ShellCheck/Parser.hs
```diff
@@ -150,13 +182,15 @@ data HereDocContext =
 data UserState = UserState {
     lastId :: Id,
     positionMap :: Map.Map Id SourcePos,
```
koalaman: maybe this should be `Map Id (SourcePos, Maybe SourcePos)` with a goal of removing the Maybe at some later point, rather than adding a separate map?

ShellCheck/Parser.hs
eatnumber1: Much nicer. Done

ngzhian:
I would like to pick this up :) Working on it locally to fix the merge conflicts, and also make it a `Map Id (SourcePos, Maybe SourcePos)`.

koalaman:
This was intended to allow fix suggestions once upon a time. https://github.com/anordal/shellharden has since aimed to solve a similar problem.

I don't know how much effort it would be to get this working consistently, and whether it's worth it.

ngzhian:
With #1239 merged, I think we can close this :)

PR #720, opened 2016-08-01
Dockerise the shellcheck

soulmachine:
This Dockerfile wraps the shellcheck as a single executable,  which is more friendly to for users

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,12 +1,14 @@
 FROM ubuntu:16.04
 MAINTAINER koalaman <koalaman@gmail.com>
 
-RUN apt-get update -qqy
+RUN apt-get update -qy
 
-RUN apt-get -qqy install haskell-platform
+RUN apt-get -qy install git cabal-install
 
-RUN cabal update
-RUN cabal install shellcheck
+RUN cd $HOME && git clone https://github.com/koalaman/shellcheck.git
+RUN cd $HOME/shellcheck && cabal update
+RUN cd $HOME/shellcheck && cabal install
+
 
 # Clean apt cache
 RUN apt-get clean && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,20 +1,16 @@
 FROM ubuntu:16.04
 MAINTAINER koalaman <koalaman@gmail.com>
 
-RUN apt-get update -qy
-
-RUN apt-get -qy install git cabal-install
-
-RUN cd $HOME && git clone https://github.com/koalaman/shellcheck.git
-RUN cd $HOME/shellcheck && cabal update
-RUN cd $HOME/shellcheck && cabal install
-
-
-# Clean apt cache
-RUN apt-get clean && \
+RUN apt-get update -y && \
+    apt-get -y --no-install-recommends install git cabal-install && \
+    apt-get clean && \
     apt-get autoremove && \
-    rm -rf /var/lib/apt/lists/* \
-    rm -rf /var/cache/*.tar.gz \
-    rm -rf /var/cache/*/*.tar.gz
+    rm -rf /var/lib/apt/lists/*
+
+RUN git config --global http.sslVerify false
+WORKDIR /root
+RUN git clone https://github.com/koalaman/shellcheck.git
+WORKDIR /root/shellcheck
+RUN cabal update && cabal install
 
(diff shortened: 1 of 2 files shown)


brother:
When is it practically useful to execute a full ubuntu docker instance instead of installing the shellcheck locally?

Adding a docker instance for the web service would be a thing. Adding this "shortcut" (not that handy imho...) into that instance would be perfectly fine.

soulmachine:
To install shellcheck you need to install the huge haskell platform, not all users have the root privilege  to install packages to the OS.

brother:
Sure local package install is a thing. But does that motivate a 2 GB of docker instance? Maybe it does. Maybe those persons could do with shellcheck.net instead? 

Having a docker instance for the shellcheck service is still a good thing imho.

When focusing on the use case of not being able to install system packages another approach could be to provide snap (and/or other format) package for shellcheck.

koalaman:
Docker sounds like a good idea to me. This Dockerfile currently installs the latest version from Hackage, though. It ignores any changes you make to the source directory.

How can it be changed to build and dockerize the current directory instead?

soulmachine:
@koalaman `apt-get install shellcheck` will install an much old version than `cabal install shellcheck`

koalaman:
Sure, but let's say we wanted the latest commit so you can make a change and build a docker image for it

soulmachine:
Got it, I'll modify the `Dockerfile` soon

soulmachine:
@kpankonen Your Dockerfile can not run at all, the error message is `Step 4 : COPY ShellCheck.cabal /src/ShellCheck.cabal
lstat ShellCheck.cabal: no such file or directory`, besides, why make a duplicated PR ?

kpankonen:
@soulmachine it's meant to be run from inside the shellcheck repo.  It lets you build with any local changes that you've made.  Your latest version will only build whatever is in master in koalman/shellcheck (instead of your local version.)

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu:16.04
+MAINTAINER koalaman <koalaman@gmail.com>
+
+RUN apt-get update -qy
```
kpankonen: apt-get update and apt-get install should be combined in the same run statement.  See https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/apt-get

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get -qy install git cabal-install
+
+RUN cd $HOME && git clone https://github.com/koalaman/shellcheck.git
```
kpankonen: This is going to result in a cached layer, so you will never get an updated version of the repository unless you invalidate the intermediate layer

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN cd $HOME/shellcheck && cabal update
+RUN cd $HOME/shellcheck && cabal install
+
+
+# Clean apt cache
+RUN apt-get clean && \
```
kpankonen: You need to run these cleanup tasks in the same RUN statement as your update/install.  Otherwise you'll have a layer with all of these files, and then another layer that removes them.  You won't be getting the space savings that you expect.

kpankonen:
@soulmachine you might want to try a Dockerfile linter like [hadolint](https://github.com/lukasmartinelli/hadolint)

docker/Dockerfile
soulmachine: Good point, I agree that `apt-get update`  should be combined with `apt-get run`

docker/Dockerfile
```diff
@@ -1,12 +1,14 @@
 FROM ubuntu:16.04
 MAINTAINER koalaman <koalaman@gmail.com>
 
-RUN apt-get update -qqy
+RUN apt-get update -qy
 
-RUN apt-get -qqy install haskell-platform
+RUN apt-get -qy install git cabal-install
+
+RUN cd $HOME && git clone https://github.com/koalaman/shellcheck.git
```
koalaman: Doesn't this clone a new copy from github rather than using the source code you currently have and may have modified?

soulmachine: Yes, this Dockerfile clones from github directly

koalaman:
I merged the other PR that builds from the current source instead of from github. Thanks for the suggestion!

PR #722, opened 2016-08-05
add Dockerfile that will build shellcheck

kpankonen:
Similar to #720 but will build from the current source.  Run with docker run -i shellcheck < /path/to/script

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM ubuntu:xenial
+MAINTAINER https://github.com/koalaman/shellcheck
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    cabal-install \
+    ghc \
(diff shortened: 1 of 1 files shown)


koalaman:
Neat. 

This results in a 1.357 GB image that presumably contains the entire Haskell build stack. The ShellCheck binary itself is ~20MB and doesn't need any Haskell libraries or tools. 

Is there a way to get rid of the unnecessary gigabyte?

kpankonen:
https://github.com/kpankonen/shellcheck/blob/docker2/Dockerfile reduces the size of the image to 330.8 MB.  

It's a trade-off with build time though.  The first Dockerfile will cache the build (apt and Haskell) dependencies, so you can quickly create a new build after changing the source.  The second Dockerfile would require you to download/build all the dependencies on every build.

PR #774, opened 2016-11-11
Push git tags to dockerhub as docker tags

oyvindio:
It is nice that shellcheck has a Docker distribution now since #734 was merged. It is useful in some environments, for instance when using shellcheck in Travis builds.
As far as I can understand, it looks like the current setup will only push the latest build of HEAD of the master branch to the `latest` tag in Dockerhub though. I think it would be useful if the releases were also tagged in Dockerhub so that you can point to a release version when using the Docker distribution of shellcheck e.g. `docker run koalaman/shellcheck:v0.4.5`.

This change *should* do just that, but it is a bit difficult to test. :) 
It looks like Travis will set `$TRAVIS_TAG='git_tag'` and `$TRAVIS_BRANCH='git_tag'` if the build was initiated by pushing a tag to the git repo. (see https://docs.travis-ci.com/user/environment-variables/) So the rationale is that we check for both of those being set to to the same thing and push `$TAG` to dockerhub if that is the case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ services:
 before_install:
   - export DOCKER_REPO=koalaman/shellcheck
   - |-
-    export TAG=$([ "$TRAVIS_BRANCH" == "master" ] && echo "latest" || echo $TRAVIS_BRANCH)
+    export TAG=$([ "$TRAVIS_BRANCH" == "master" ] && echo "latest" || ([ -n "$TRAVIS_TAG" ] && echo "$TRAVIS_TAG") || echo "$TRAVIS_BRANCH")
 
 script:
   - docker build -t builder -f Dockerfile_builder .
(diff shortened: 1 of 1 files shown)


koalaman:
Skjera :D 

How would it do that? Isn't `$TAG` the variable set earlier in before_install? Should it be `[ -n "$TRAVIS_TAG" ] && docker push "$DOCKER_REPO:$TRAVIS_TAG"` or something?

oyvindio:
> Skjera :D
> 

Hei :)

> How would it do that? Isn't $TAG the variable set earlier in before_install? Should it be [ -n "$TRAVIS_TAG" ] && docker push "$DOCKER_REPO:$TRAVIS_TAG" or something?

Yes, but `$TAG` is either set to `$TRAVIS_BRANCH` or `"latest"` if we are on master, and it seems that Travis sets `$TRAVIS_BRANCH="the_git_tag"` when building a tag. It would be better to depend on `$TRAVIS_TAG` though, since that's what the documentation says to use.

So how about setting `$TAG` to "latest" if we are on master, else `$TRAVIS_TAG` if defined, else `$TRAVIS_BRANCH`. Then we can essentially do `([ $TRAVIS_BRANCH == "master"] || [ -n "$TRAVIS_TAG" ]) && docker push "$DOCKER_REPO:$TAG"` in the `after_success` section.

koalaman:
The logic is somewhat hairy but it should be ok for now

Issue #562, opened 2015-12-15
SC2140 triggers for slash on parameter expansion replacement


cdown:
Shellcheck complains that `/` is not quoted, but it's part of the expansion syntax, it's not something to be quoted.

```
% shellcheck --version
ShellCheck - shell script analysis tool
version: 0.4.1
license: GNU General Public License, version 3
website: http://www.shellcheck.net
% shellcheck -s bash /dev/stdin << 'EOF'
>>> : "${PHOTO_DIR/"'"/"'\\''"}"
>>> EOF                         

In /dev/stdin line 1:
: "${PHOTO_DIR/"'"/"'\\''"}"
                  ^-- SC2140: Word is on the form "A"B"C" (B indicated). Did you mean "ABC" or "A\"B\"C"?

```

MortimerMcMire315:
I think the root issue here is that the quoted parts in the expansion are being parsed as `T_NormalWord`, and therefore (as they should) match the first pattern in `checkInexplicablyUnquoted`. As far as I can see, ShellCheck does not recognize parameter substitution. So the "right" way to do this would be to add a parser for parameter substitution. I hope to have this working on my fork sometime in the next few days.

As a side note, I'm thinking the error message is meant to say "Word is *of* the form" (rather than "on").

PR #786, opened 2016-11-28
Add handling for special characters in parameter substitutions.

MortimerMcMire315:
Fixes koalaman/shellcheck#562. Special characters inside braces are
parsed into T_ParamSubSpecialChar instead of T_Literal so that they are
not flagged in the function checkInexplicablyUnquoted when sandwiched
between double quotes.


koalaman:
Thanks!

PR #814, opened 2017-01-02
This commit fixes #797

### LINKED ISSUE #797: [SC2126] doesn't check for wc's parameters
#### Bug
- Rule Id: SC2126 
- My shellcheck version: 0.3.7
- I tried on shellcheck.net and verified that this is still a problem on the latest commit

```sh

#!/bin/sh
echo "something else" | grep "some" | wc -w

```

SC2126: Consider using grep -c instead of grep|wc.

nothing in this case since wc -c cannot be worked around with grep.
It should check for wc's parameters and if it has "-l" but doesn't contain "-c", "-m", or "-w", then it should print:
SC2126: Consider using grep -c instead of grep | wc -l.
else print nothing

EDIT: formatting

Scorpiokat:
Added check for wc's params when used with grep as has been suggested by the issue reporter. Also added a range of tests.


koalaman:
Thanks!

PR #822, opened 2017-01-10
SC2164: show two possible variants for circumenting the warning

bittorf:
always calling 'exit' is not good in e.g. functions.
the basic idea is at least that the returncode of
cd *is* evaluated somehow and not ignored.

Reported-by: Garance Alistair Drosehn <drosehn@rpi.edu>

Signed-off-by: Bastian Bittorf <bittorf@bluebottle.com>


koalaman:
Thanks!

PR #860, opened 2017-03-27
Update README.md

xenopeek:
Command to install on Arch Linux based distros


koalaman:
Always the first distro to update to new releases. Thanks!

PR #903, opened 2017-05-14
SC2164: Make SC2164 apply to `pushd` and `popd`

### LINKED ISSUE #863: `pushd` and `popd` should be treated like `cd`
#### For bugs
- Rule Id (if any, e.g. SC1000): 
- My shellcheck version (`shellcheck --version` or "online"): 0.4.6

```sh
#!/bin/bash
cd /foo
pushd /bar
popd
```

```
In demo.sh line 2:
cd /foo
^-- SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
```

```
In demo.sh line 2:
cd /foo
^-- SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.


In demo.sh line 3:
pushd /bar
^-- SC2164: Use 'pushd ... || exit' or 'pushd ... || return' in case pushd fails.


In demo.sh line 4:
popd
^-- SC2164: Use 'popd || exit' or 'popd || return' in case popd fails.
```
`pushd` and `popd` have exactly the same failure cases as `cd`; SC2164 should apply to them as well.

mrshu:
* Since `pushd` and `popd` have the same failure cases, make the check
  for SC2164 apply to them as well.

* This commit also refactors the code a bit as `hasSetE` is now used in
  multiple places.

* Fixes #863.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/ShellCheck/Analytics.hs
+++ b/ShellCheck/Analytics.hs
@@ -62,6 +62,8 @@ treeChecks = [
     ,checkShebang
     ,checkUnassignedReferences
     ,checkUncheckedCd
+    ,checkUncheckedPushd
+    ,checkUncheckedPopd
     ,checkArrayAssignmentIndices
     ]
 
@@ -2539,7 +2541,7 @@ prop_checkUncheckedCd6 = verifyNotTree checkUncheckedCd "cd .."
 prop_checkUncheckedCd7 = verifyNotTree checkUncheckedCd "#!/bin/bash -e\ncd foo\nrm bar"
 prop_checkUncheckedCd8 = verifyNotTree checkUncheckedCd "set -o errexit; cd foo; rm bar"
 checkUncheckedCd params root =
-    if hasSetE then [] else execWriter $ doAnalysis checkElement root
+    if hasSetE root then [] else execWriter $ doAnalysis checkElement root
   where
     checkElement t@T_SimpleCommand {} =
         when(t `isUnqualifiedCommand` "cd"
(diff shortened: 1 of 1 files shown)


ShellCheck/Analytics.hs
```diff
@@ -2548,7 +2550,47 @@ checkUncheckedCd params root =
                 warn (getId t) 2164 "Use 'cd ... || exit' or 'cd ... || return' in case cd fails."
```
koalaman: Neat! How about adding pushd/popd to the checkUncheckedCd method to avoid reiterating the tree several times. Maybe something like 

    checkElement t@T_SimpleCommand {} =
      when (name `elem` ["cd", "pushd", "popd"])
          ...  $ do
          warn (getId t) 2164 $ "Use '" ++ name ++ " ... || exit'  ...."
      where
        name = fromMaybe "" $ getCommandName t
        isSafeDir t = case oversimplify t of 
              [_, ".."] -> True; 
              _ -> False

mrshu:
@koalaman thanks for your suggestion -- it was indeed a very good one! Also many thanks for putting `hasSetE` into the lib -- I believe that is where it should indeed be placed.

Please let me know what you think!

ShellCheck/Analytics.hs
```diff
@@ -2463,26 +2463,48 @@ checkReadWithoutR _ t@T_SimpleCommand {} | t `isUnqualifiedCommand` "read" =
...
+prop_checkUncheckedPopd1 = verifyTree checkUncheckedCdPushdPopd "popd; rm -r foo"
+prop_checkUncheckedPopd2 = verifyNotTree checkUncheckedCdPushdPopd "popd || exit; rm -r foo"
+prop_checkUncheckedPopd3 = verifyNotTree checkUncheckedCdPushdPopd "set -e; popd; rm -r foo"
+prop_checkUncheckedPopd4 = verifyNotTree checkUncheckedCdPushdPopd "if popd; then rm foo; fi"
+prop_checkUncheckedPopd5 = verifyTree checkUncheckedCdPushdPopd "if true; then popd; fi"
+prop_checkUncheckedPopd6 = verifyTree checkUncheckedCdPushdPopd "popd"
+prop_checkUncheckedPopd7 = verifyNotTree checkUncheckedCdPushdPopd "#!/bin/bash -e\npopd\nrm bar"
+prop_checkUncheckedPopd8 = verifyNotTree checkUncheckedCdPushdPopd "set -o errexit; popd; rm bar"
+
+checkUncheckedCdPushdPopd params root =
```
mrshu: @koalaman I am not sure if you agree with the change of the name of the function, but it did feel appropriate to me.

ShellCheck/Analytics.hs
```diff
@@ -2463,26 +2463,48 @@ checkReadWithoutR _ t@T_SimpleCommand {} | t `isUnqualifiedCommand` "read" =
...
+    if hasSetE params then
+        []
     else execWriter $ doAnalysis checkElement root
   where
     checkElement t@T_SimpleCommand {} =
-        when(t `isUnqualifiedCommand` "cd"
-            && not (isCdDotDot t)
+        when(name t `elem` ["cd", "pushd", "popd"]
+            && not (isSafeDir t)
+            && not (name t == "pushd" && ("n" `elem` map snd (getAllFlags t)))
```
mrshu: @koalaman I do not really like this line, but sadly, this is the only way I managed to get `pushd -n` to be skipped.

If you can think of a better way, please let me know.

koalaman:
Thanks!

PR #904, opened 2017-05-15
SC2174: Do not warn at `mkdir -pm 0700 ../foo`

### LINKED ISSUE #854: SC2174: should not warn about mkdir -p -m 0755 ../foo
In a shell script I am using
```
#!/bin/sh
mkdir -p -m 0711 ../bin
```
to make sure that the directory `bin` exists with the right permissions. I use `-p` to make sure `mkdir` does not give an error when the directory already exists.

Shellcheck version 0.4.4 generates a warning here:
```
../judge/testcase_run.sh:172:10: warning: When used with -p, -m only applies to the deepest directory. [SC2174]
```
Although I realize this is a corner case, I think that any composite path where all but the last element consists of `.` or `..` should not trigger this warning.

mrshu:
* Do not warn when `mkdir -pm 0700` is used with combination of paths
  like `..` and `.`

* Fixes #854

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/ShellCheck/Checks/Commands.hs
+++ b/ShellCheck/Checks/Commands.hs
@@ -385,17 +385,26 @@ prop_checkMkdirDashPM11 = verifyNot checkMkdirDashPM "mkdir --parents a/b"
 prop_checkMkdirDashPM12 = verifyNot checkMkdirDashPM "mkdir --mode=0755 a/b"
 prop_checkMkdirDashPM13 = verifyNot checkMkdirDashPM "mkdir_func -pm 0755 a/b"
 prop_checkMkdirDashPM14 = verifyNot checkMkdirDashPM "mkdir -p -m 0755 singlelevel"
+prop_checkMkdirDashPM15 = verifyNot checkMkdirDashPM "mkdir -p -m 0755 ../bin"
+prop_checkMkdirDashPM16 = verify checkMkdirDashPM "mkdir -p -m 0755 ../bin/laden"
+prop_checkMkdirDashPM17 = verifyNot checkMkdirDashPM "mkdir -p -m 0755 ./bin"
+prop_checkMkdirDashPM18 = verify checkMkdirDashPM "mkdir -p -m 0755 ./bin/laden"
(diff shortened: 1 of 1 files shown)


ShellCheck/Checks/Commands.hs
```diff
@@ -385,17 +385,25 @@ prop_checkMkdirDashPM11 = verifyNot checkMkdirDashPM "mkdir --parents a/b"
...
+        -- mkdir -pm 0700 dir  is fine, ../dir may be too, but dir/subdir is not.
+        guard $ any couldHaveSubdirs (drop 1 $ arguments t)
         return $ warn (getId $ fst dashM) 2174 "When used with -p, -m only applies to the deepest directory."
     couldHaveSubdirs t = fromMaybe True $ do
         name <- getLiteralString t
-        return $ '/' `elem` name
+        return $ '/' `elem` name &&
+                not((".." `isInfixOf` name || "." `isInfixOf` name) &&
+                    countLetters name '/' == 1)
+    countLetters str c = length $ filter (== c) str
```
mrshu: @koalaman I am not sure if this is the correct approach -- feel free to let me know if it should be done differently.

eldering:
I did a quick check, and it indeed ignores prefix `./` or `../` but not longer variants, such as `../../bin`. I think a prefix of any length consisting of `.` and `..` components should be ignored.

mrshu:
@eldering Thanks for your comment @eldering, I refactored the PR to account for that too. Feel free to check it again, a prefix of any length consisting of `.` and `..` components should be ignored.

eldering:
@mrshu Thanks, tested it and it looks good to me.

koalaman:
Nice, thanks!

kurahaupo:
If `-p` is present, it simply should not check for `.` or `..` at all, as it
is valid to say:

```
    mkdir -p ../foo/../bar/..
```

This has the effect of creating both ../foo & ../bar if they do not already
exist, or if ../foo is a symlink to an existing directory, creating bar in
the parent of that directory (if it does not exist).

A bit esoteric, but not an error.

However as noted elsewhere, supplying the -m option only applies to the last
component, and only if it is created.

koalaman:
O_O

PR #912, opened 2017-05-22
Recognize bash's `shopt -s lastpipe`

### LINKED ISSUE #732: `shopt -s lastpipe` should disable SC2030/SC2031 appropriately
ShellCheck correctly warns on this snippet (from the SC2031 wiki):

``` bash
n=0
printf "%s\n" {1..10} | while read i; do (( n+=i )); done
echo $n
```

However, the code will behave as expected if `shopt -s lastpipe` is used, making the warning spurious.

Perhaps ShellCheck should detect this option is in use and stop treating the last command in a pipeline as a subshell, in the same way that `set -e` disables `cd` warnings.

CyberShadow:
Fixes #732.


CyberShadow:
Currently things like `set -e` seem to be implemented as a hack; `set +e` is not recognized, and nor is using `set -e` *after* `cd`, because ShellCheck does not keep track of what shell options are set at any given moment. The same applies to this fix, as it uses the same method used to detect `set -e`.

koalaman:
Nice, thanks!

Issue #950, opened 2017-07-12
Incorrect parsing of default value in variable expansion


martin-schwenke:
#### For bugs
- Results in SC2031, but the problem isn't really with the rule
- 0.4.6
- I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/sh
b=$(d=$(dirname "$0") ; cd -P "$d" || exit ; dirname "$PWD")
bar="${FOOBAR:-${b}/foobar.d}"
echo "$bar"

```

#### Here's what shellcheck currently says:

In foo.sh line 2:
b=$(d=$(dirname "$0") ; cd -P "$d" || exit ; dirname "$PWD")
    ^-- SC2030: Modification of d is local (to subshell caused by $(..) expansion).

In foo.sh line 3:
bar="${FOOBAR:-${b}/foobar.d}"
       ^-- SC2031: d was modified in a subshell. That change might be lost.

#### Here's what I wanted or expected to see:

Nothing!

"d" is erroneously parsed as a variable name when it is part of a default value string.  This also happens if 'd' is preceded by '-' or '+' but not '_'.  So it looks like the default value is not treated at a string.  It looks like the tail end of it is treated as an identifier.

Thanks...

martin-schwenke:
This didn't happen in 0.4.4, so I've hand bisected and found the commit that appears to introduce the regression:

a90b6d1 Count b as a reference in ${a:b}

Here's the difference between that commit and the previous one:

```
$ git reset --hard a90b6d1
HEAD is now at a90b6d1 Count b as a reference in ${a:b}
$ cabal build
Building ShellCheck-0.4.4...
Preprocessing library ShellCheck-0.4.4...
[11 of 17] Compiling ShellCheck.ASTLib ( ShellCheck/ASTLib.hs, dist/build/ShellCheck/ASTLib.o )
[12 of 17] Compiling ShellCheck.Parser ( ShellCheck/Parser.hs, dist/build/ShellCheck/Parser.o )
[13 of 17] Compiling ShellCheck.AnalyzerLib ( ShellCheck/AnalyzerLib.hs, dist/build/ShellCheck/AnalyzerLib.o )
[14 of 17] Compiling ShellCheck.Analytics ( ShellCheck/Analytics.hs, dist/build/ShellCheck/Analytics.o )
[...]

PR #957, opened 2017-07-20
Fix incorrect detection of bash-style substring expansion (issue #950)

martin-schwenke:
Substring expansion detection only considers ':' as a separator..  It
needs to avoid triggering for ":-", ":=", ":+" and ":?", since they
mean other things.

This is a regression introduced by commit
a90b6d14b3de0f21db462afbc925444687feb2ca

Signed-off-by: Martin Schwenke <martin@meltin.net>


koalaman:
Thanks!

PR #968, opened 2017-08-11
Documented binary cabal install for MacOS

ssbarnea:
Installing haskell binaries is 50x faster than compiling it from source.


koalaman:
Thanks!

Issue #953, opened 2017-07-13
discrepancy between `declare -A` and `local -A`


mmlb:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2154
- My shellcheck version (`shellcheck --version` or "online"): online
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/usr/bin/bash

function arrDeclare() {
  declare -A arrd
  arrd[foo]=bar
  echo "${arrd[*]}"
}

function arrLocal() {
  local -A arrl
  arrl[fool]=bar
  echo "${arrl[*]}"
}

```

#### Here's what shellcheck currently says:

```
Line 11:
  arrl[fool]=bar
       ^-- SC2154: fool is referenced but not assigned.
```

#### Here's what I wanted or expected to see:
nothing

#### More info
This is similar to #359, the difference is in how the array was declared. According to [bash docs](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-local) `local` is synonymous with `declare` + local to function.

Nightfirecat:
I'm able to reproduce the same in my script; changing `local` to `declare` stops SC2154 from triggering.

PX03AFK:
Also reported in #958

mmlb:
yay :tada:

PR #988, opened 2017-09-03
SC2154: Fix false positive on `local`

Nightfirecat:
Test file:

```
$ cat test.sh
#!/usr/bin/bash

function arrDeclare() {
  declare -A arrd
  arrd[foo]=bar
  echo "${arrd[*]}"
}
function arrLocal() {
  local -A arrl
  arrl[fool]=bar
  echo "${arrl[*]}"
}
function arrTypeset() {
  typeset -A arrt
  arrt[foot]=bar
  echo "${arrt[*]}"
}
```

Before this fix:

```
$shellcheck test.sh

In test.sh line 10:
  arrl[fool]=bar
       ^-- SC2154: fool is referenced but not assigned.
```

After this fix:

```
$shellcheck test.sh
```

Fixes koalaman/shellcheck#953

P.S. I'm not very familiar with the test frameworks or with Haskell, so I did not add any tests for these additional cases. I'll happily accept guidance to where and how that should be done!

### DIFF
--- a/ShellCheck/ASTLib.hs
+++ b/ShellCheck/ASTLib.hs
@@ -329,7 +329,8 @@ getAssociativeArrays t =
     f :: Token -> Writer [String] ()
     f t@T_SimpleCommand {} = fromMaybe (return ()) $ do
         name <- getCommandName t
-        guard $ name == "declare" || name == "typeset"
+        let assocNames = ["declare","local","typeset"]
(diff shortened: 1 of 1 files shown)


Nightfirecat:
D'oh, had title and commit message wrong; fixing those now.

koalaman:
Thanks!

PR #992, opened 2017-09-06
readme: update supported languages

mcandre:
update readme to reflect the languages actually supported by shellcheck


koalaman:
ShellCheck doesn't support `ash` at this time. Specifying `ash` only switches to `dash`, since that's at least a bit more correct than defaulting to `bash`.

PR #1032, opened 2017-10-19
Add a pre-commit config so shellcheck can be used as a pre-commit hook

jesselang:
I'd like to use shellcheck as a pre-commit hook for some shell script projects. To use this hook in a repo, create a `.pre-commit-config.yaml` in the repo root with the following, then run `pre-commit install --install-hooks`.

```
repos:
  - repo: https://github.com/koalaman/shellcheck
    sha: v0.4.7 # or whatever tag this change is release in
    hooks:
      - id: shellcheck
```

Seems to work great!

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+- id: shellcheck
+  name: shellcheck
+  description: |
+    A static analysis tool for shell scripts
+    https://www.shellcheck.net
+  entry: koalaman/shellcheck
+  language: docker_image
+  types: [shell]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,6 +3,15 @@
   description: |
     A static analysis tool for shell scripts
     https://www.shellcheck.net
-  entry: koalaman/shellcheck
+  entry: koalaman/shellcheck:stable
+  language: docker_image
+  types: [shell]
+
+- id: shellcheck_latest
+  name: shellcheck-latest
+  description: |
+    A static analysis tool for shell scripts - latest tag
+    https://www.shellcheck.net
+  entry: koalaman/shellcheck:latest
   language: docker_image
   types: [shell]


jesselang:
Hey @koalaman. Any thoughts on this PR?

koalaman:
Sounds great! 

I'm not familiar with this system though. The docker image `koalaman/shellcheck` defaults to the latest commit, so it seems like this file would always check against the latest commit rather than the version specified. How can we get it to use the associated stable version instead?

Sorry about the slow response.

jesselang:
I've updated to use the `stable` tag. Thanks!

jesselang:
I've also added a separate hook so those who want to can take advantage of the `latest` tag.

koalaman:
So would you now not specify `sha` since you can't specify a shellcheck version?

jesselang:
The repo using `pre-commit` that consumes the `shellcheck` may choose to specify a `sha` as part of their usage of the hook. To better facilitate that model of use, perhaps the hook that currently uses `stable` should use a deterministic tag, such as `v0.4.6`, and the tag in the `pre-commit-hooks.yaml` file could be incremented in the commit that gets tagged with the next version.

jesselang:
Any more thoughts here, @koalaman?

koalaman:
I'm currently trying to avoid having to update version numbers in multiple places. Is there an alternative?

KingRando95:
Pull request

jesselang:
Not to my knowledge. I'll close this PR for now. Feel free to reopen or otherwise adapt the code at a later time. Thanks!

matthewpersico:
bumpversion? (https://pypi.org/project/bumpversion/)

SimonHeimberg:
I suggest to use tag stable for docker only.

anybody wanting a different version can set this in the `.pre-commit-config.yml`:
```yaml
repos:
  - repo: https://github.com/koalaman/shellcheck
    sha: v0.4.7 # or whatever tag this change is release in
    hooks:
      - id: shellcheck
        entry: koalaman/shellcheck:v0.4.6
```
##### another suggestion:
only provide the "normal" hook:
```yaml
- id: shellcheck
  name: shellcheck
  description: |
    A static analysis tool for shell scripts
    https://www.shellcheck.net
  language: system
  entry: shellcheck
  args: [--exclude, SC1091] # do not fail when source script is not checked. or use [--external-sources]
  types: [shell]
  additional_dependencies: [shellcheck]
```
args and additional_dependencies picked from: https://github.com/jumanjihouse/pre-commit-hooks/blob/01cbe08/.pre-commit-hooks.yaml#L68
another existing .pre-commit-hooks.yaml: https://github.com/detailyang/pre-commit-shell/blob/master/.pre-commit-hooks.yaml


anybody wanting to run in docker can set this in the `.pre-commit-config.yml`:
```yaml
repos:
  - repo: https://github.com/koalaman/shellcheck
    sha: v0.4.7 # or whatever tag this change is release in
    hooks:
      - id: shellcheck
        language: docker_image
        entry: koalaman/shellcheck:v0.4.6 # or latest, ...
```
Setting a different language is not officially supported, so I'm not sure if this a good idea.

Issue #862, opened 2017-03-28
SC2034 Interprets `{var}>&-` and `{var}<&-` redirections as variable assignment instead of access


LukeShu:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2034
- My shellcheck version (`shellcheck --version` or "online"): 0.4.6
- [X] I tried on shellcheck.net and verified that this is still a problem on the latest commit
- [ ] It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/bash
fd_close() {
    local fd=$1
    exec {fd}>&-
}
```

#### Here's what shellcheck currently says:

```
Line 4:
        exec {fd}>&-
             ^-- SC2034: fd appears unused. Verify it or export it.
```

#### Here's what I wanted or expected to see:

Nothing.

nihilus:
Shouldnt it be '${fd}'?

LukeShu:
No, the body of the bug report says exactly what I meant for it to say.  (I have, however, changed the title to be more descriptive of the problem).

  If I had written what @nihilus suggested,

    exec ${fd}>&-

the shell would interpret the value of `${fd}` as a simple command, rather than as the file descriptor number to operate on.  To achieve the desired effect with "normal" variable expansion, we'd have to use eval:

    eval "exec $fd>&-"

However, this isn't necessary, because Bash 4.1 introduced the simpler syntax that this bug is about.  To quote the Bash 4.1 release notes:

    p.  If the optional left-hand-side of a redirection is of the form {var}, the
        shell assigns the file descriptor used to $var or uses $var as the file
        descriptor to move or close, depending on the redirection operator.

The feature is explained in the Bash manual:

> Each redirection that may be preceded by a file descriptor number may instead be preceded by a word of the form {varname}.  In this case, for each redirection operator except >&- and <&-, the shell will allocate a file descriptor greater than or equal to 10 and assign it to varname.  If >&- or <&- is preceded by {varname}, the value of varname defines the file descriptor to close.

nihilus:
Aha, so this is some kinda arcane bashism... I see.

Nightfirecat:
@koalaman This should be closed. (#1041 probably should have said `fixes #862` to directly close it)

LukeShu:
Huh, I'm surprised that merging #1041 didn't automatically close this.  I guess sticking "issue" in there broke it.

[...]

PR #1041, opened 2017-11-06
Fix isClosingFileOp (fixes issue #862)

LukeShu:
The isClosingFileOp function expected closing file ops to use T_IoFile, but
they actually use T_IoDuplicate; so it effectively always returned False.

This fixes issue #862.


koalaman:
Very nice, thanks!

PR #1093, opened 2018-01-12
Add snapcraft.yaml to build snap package

albertodonato:
This adds configuration to build a snap package (https://snapcraft.io/)

To build, run `snapcraft` from the base directory.
The resulting snap can be installed with `sudo snap install --dangerous --classic <snapfile>`

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,37 @@
+name: shellcheck
+summary: A shell script static analysis tool
+description: |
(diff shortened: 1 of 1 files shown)


koalaman:
I have no idea how snap works, so I'll take your word for it. Thanks!

albertodonato:
@koalaman thanks for merging the PR. fwiw you could now go to build.snapcraft.io and enable snap build/publish so that it can be installed from the store (with `snap install --classic shellcheck`).

PR #1100, opened 2018-01-24
SC2029: Skip when there are options to ssh

### LINKED ISSUE #327: SC2029 false positive
SC2029 is reported for this line:

```
ssh -f -N -R 60000:localhost:60000 "user@$ip"
```

I suppose SC considers `"user@$ip"` something that's executed on the remote server, but it's not.

mkhl:
Fixes #327

SC2029 generates false positives when given an ssh command that includes
options with arguments because it assumes the first non-option must be the
host:port argument and the last argument is a command to run.
As suggested the comments on #327, this change fixes those by skipping the
check when there are any options present.


koalaman:
Thanks!

PR #1112, opened 2018-02-08
Add test for `local` keyword in SC2154

Nightfirecat:
I felt bad about not adding a test for this in my original PR, so this should cover that behavior.

Follow-up to koalaman/shellcheck#988


koalaman:
Thanks!

PR #1123, opened 2018-02-22
Check for calls to `which`.

jonhiggs:
Favour the builtin `command -v` instead.

Rationale is discussed at https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script.


koalaman:
Looks good to me!

`./nextnumber` on head says the next available number is now 2228, could you update it?

jonhiggs:
All done.

koalaman:
Thanks!

PR #1140, opened 2018-03-08
Move library into src/

phadej:
This speedups builds as library part is built only once

```diff
-cabal new-build --enable-tests  395,61s user 6,54s system  97% cpu 6:51,35 total
+cabal new-build --enable-tests  165,86s user 5,82s system 100% cpu 2:51,07 total
```

(#1139 is about first commit).


koalaman:
Thanks!

I'm not familiar with Cabal at all, so I really appreciate PRs like this.

PR #1143, opened 2018-03-14
Simplify Dockerfile

pratikmallya:
Use multi stage Dockerfile to greatly simplify build.

PR #1151, opened 2018-03-21
Fix Docker images

shaneog:
#1143 broke the functioning of the Docker images.

This PR restores that functionality for both Docker image types.


koalaman:
Thanks! This was not caught by the tests that didn't actually try to check a file, only invoke the binary.

I've added a test for this and updated the workdir of the koalaman/shellcheck image in 795a8812 like you suggested. I left shellcheck-alpine alone though, because it's just intended to be alpine with shellcheck preinstalled, and not have shellcheck as its entry point.

PR #1159, opened 2018-03-25
Consider type a valid command in sh

geirha:
[type](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/type.html) *is* defined by POSIX, and even the bourne shell has the `type` command.


koalaman:
Thanks!

PR #1168, opened 2018-04-02
Bump to ghc 8.4.1

vmchale:
Currently, shellcheck fails to build with ghc-8.4.1. This fixes that, while maintaining compatibility with ghc-7.8.4.


koalaman:
Thanks!

Issue #1137, opened 2018-03-06
Warn about `sudo cd`,`sudo export` and similar


koalaman:
#### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/sh
sudo cd /root

```

#### Here's what shellcheck currently says:

Nothing

#### Here's what I wanted or expected to see:

Can't use `sudo` with `cd`

PR #1195, opened 2018-04-29
Warn about invalid arguments to sudo

sdknudsen:
Addresses https://github.com/koalaman/shellcheck/issues/1137


koalaman:
Thanks!

PR #1203, opened 2018-05-06
Whitelist rename for SC2016 (fixes #1199)

### LINKED ISSUE #1199: SC2016 gives false positive on regex parameters
#### For bugs
- Rule Id:
SC2016
- version:
0.3.7
- I tried on shellcheck.net and verified that this is still a problem

```sh

#!/bin/bash

rename 's/(.)a/$1/g' ./*

```

> Line 3:
> rename 's/(.)a/$1/g' ./*
>        ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

#### Notes

rename has a false positive but perl does not;

```sh

#!/bin/bash

perl -pi -e 's/(.)a/$1/g' ./*


```

ngzhian:
Add rename [0] to the list of commands that we don't check single quoted variables.

[0] http://man7.org/linux/man-pages/man1/rename.1.html


koalaman:
Thanks!

Issue #840, opened 2017-03-01
SC2181 after conditionals


arth1:
For SC2181, recommending not checking $? but instead test the command directly, it does not take into account that the return code may be based on one of several execution branches.

Consider this:

```
if some_test; then command1; else command2; fi
if [ $? -ne 0 ]; then
  # regardless of which command ran, it failed
fi
```
I can't see that this is better written as
```
if ! if some_test; then command1; else command2; fi; fi; then
  # something
fi
```

Perhaps SC2181 should not kick in directly after a conditional close like fi or esac?

kurahaupo:
On Wed, 1 Mar 2017, arth1 wrote:
> Consider this:
>
> ```
> if some_test; then command1; else command2; fi
> if [ $? -ne 0 ]; then
>   # regardless of which command ran, it failed
> fi
> ```
> I can't see that this is better written as
> ```
> if ! if some_test; then command1; else command2; fi; fi; then
>   # something
> fi
> ```

The latter definitely looks like an improvement to me (after fixing it to
remove the extra "fi", and perhaps reformatting it for better legibility:

```
if ! if some_test ; then
        command1
     else
        command2
     fi
then
  # something
fi
```

Or perhaps more succinctly:

```
if some_test
then command1
else command2
fi || handle_failure
```

Or a lot of ills can be cured by wrapping things in functions:
```
do_this_or_that() {
    if some_test
    then command1
    else command2
    fi
}


if ! do_this_or_that ; then
    handle_failure
fi
```

kurahaupo:
(I suspect half the reason that these types of constructs are regarded as awkward is because people insist -- sometimes via "style guides" -- on putting "then" on the end of the last line of the test, rather than allowing it to be on its own line when that adds clarity.)

arth1:
The ";then" is irrelevant here, and the concatenation into a single line was for brevity.
The "if ! if" or "if if" construct is awkward, no matter what kind of formatting, and even more so if the test gets split up into several lines.  Using $? is quite useful in avoiding such constructs and make the code more readable.

kurahaupo:
> The "if ! if" or "if if" construct is awkward

That is a matter of personal taste, and I disagree.

To me, the nested ifs are perfectly reasonable readable code. Whereas
inspecting $? is just ugly.

So since both ways are considered awful by someone, let's take an
alternative approach: wrap the condition block in a funcion, then we can
both be happy.

Whether that merits a different wording in the warning is an open question.

[...]

PR #1204, opened 2018-05-06
Do not warn when checking multiple exit values (fixes #1167)

### LINKED ISSUE #1167: SC2181 suggests to check exit code directly when I check for several possible values
- Rule Id: SC2181
- My shellcheck version: 0.4.7

```sh

#!/usr/bin/env bash

gzip --test "$1"
if [[ $? -eq 0 || $? -eq 2 ]]
then
    echo ok
else
    echo bad
fi
```

In shellcheck-issue.sh line 4:
if [[ $? -eq 0 || $? -eq 2 ]]
      ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.

Nothing, because the successful operation is not the only exit code I'm testing against. gzip possible exit statuses:
```
Exit status is normally 0; if an error occurs, exit status is 1. If a warning occurs, exit status is 2.
```
I want to test for both normal and warning, i.e. any of 0 and 2. And the remaining exit status - 1 falls into the else conditional path.

Moreover, shellcheck is silent for
```bash
if $? || [[ $? -eq 2 ]]
```
that is completely wrong. Since testing the first expression `$?` actually gets executed and rewrites the gzip exit code value. So that the second `[[ $? -eq 2 ]]` never matches.

P.S.
Of course, I could test just for exit status 1 in my case, but there could be new exit statues beside 0, 1, 2 in general.

ngzhian:
In the check, it goes up one level to the parent, matches on the same TC_Binary, and sees if both branches of the `||` are exit checks, if so it doesn't warn, otherwise it checks each branch.

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -2558,9 +2558,12 @@ prop_checkReturnAgainstZero6 = verifyNot checkReturnAgainstZero "[[ $R -eq 0 ]]"
 prop_checkReturnAgainstZero7 = verify checkReturnAgainstZero "(( $? == 0 ))"
 prop_checkReturnAgainstZero8 = verify checkReturnAgainstZero "(( $? ))"
 prop_checkReturnAgainstZero9 = verify checkReturnAgainstZero "(( ! $? ))"
-checkReturnAgainstZero _ token =
+prop_checkReturnAgainstZero10 = verifyNot checkReturnAgainstZero "[[ $? -eq 0 || $? -eq 2 ]]"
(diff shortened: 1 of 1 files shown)


koalaman:
I think this fix may be a bit too specific. It doesn't work with e.g. `[[ $? -eq 0 || $? -eq 4 && $retry ]]`

Ideally there should be a way to get the command that a `$?` refers to (and get the `$?`s that refer to a command). 

That would allow a more robust solution to this issue, and would be useful in several other checks like making this not trigger when the command is a long compound command (#840), the `f() { echo $?; }` case on this issue, and new checks for checking the exit code of `test` or `echo`, as in `cmd; echo "Exited with $?"; [ $? -eq 42 ] && ...`

Issue #1175, opened 2018-04-08
Conflicting advice in SC2028 and SC1117


dimo414:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC1117, SC2028
- My shellcheck version (`shellcheck --version` or 'online'): online
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit
- [ ] It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue

#### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/bash
printf "Name:\t%s\n" "$value"
```

#### Here's what shellcheck currently says:
```
Line 2	SC1117: Backslash is literal in "\t". Prefer explicit escaping: "\\t".
Line 2	SC1117: Backslash is literal in "\n". Prefer explicit escaping: "\\n".
Line 2	SC2154: value is referenced but not assigned.
```
#### Here's what I wanted or expected to see:

This line (`printf "Name:\t%s\n" "$value"`) is listed as "Correct code" in [SC2028](https://github.com/koalaman/shellcheck/wiki/SC2028), but triggers these SC1117 errors. At a glance the SC1117 errors seem wrong, but it seems that using single-quotes (`printf 'Name:\t%s\n' "$value"`) is sufficient to suppress the error, in which case the SC2028 docs should be updated.

koalaman:
You're right of course, I've updated the wiki. Additionally, the technically correct but rather noisy SC1117 has since been disabled.

PR #1205, opened 2018-05-06
Remove unused code

ngzhian:
While looking at #1175 (sorry to say I don't have a fix for it yet), i found this unused combinator. `cabal test` passes.


koalaman:
No point keeping that around. Thanks!

PR #1207, opened 2018-05-06
When given a %* format string, expect one more argument

### LINKED ISSUE #1184: SC2183: printf arguments not checking for `*` argument
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2183
- [ ] ~~It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue~~

- [x] ~~shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this~~
- [x] I searched through https://github.com/koalaman/shellcheck/issues and #1171 is not related (neither anything referencing 'printf'

```sh
#!/bin/bash

MAXIMAL=0

for E in A AB ABC ; do
	if [ "${MAXIMAL}" -lt "$(wc -c <<< "${E}")" ] ; then
		MAXIMAL="$(wc -c <<< "${E}")"
	fi
done

for E in A AB ABC ; do
	printf "%*b: %b\\n" "${MAXIMAL}" "${E}" "${E}"
done

```

```
$ shellcheck myscript
 
Line 12:
        printf "%*b: %b\\n" "${MAXIMAL}" "${E}" "${E}"
               ^-- SC2183: This format string has 2 variables, but is passed 3 arguments.
```

(Nothing). printf first string should be checked for `%*X` arguments and +1 per `*` for expected arguments

ngzhian:
Fixes #1184


koalaman:
Thanks!

PR #1208, opened 2018-05-06
Add unset to list of commands exempt from 2016

### LINKED ISSUE #1186: SC2016: Does not apply on `unset`
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2016
- [ ] ~~It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue~~

```sh
#!/bin/bash
export SPLIT=('a' 'b')
unset 'SPLIT[${#SPLIT[@]}-1]'
```

```
$ shellcheck myscript
 
Line 3:
        unset 'SPLIT[${#SPLIT[@]}-1]'
              ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

$ 
```

(Nothing). `unset` will "expand the arguments itself". Or something else happens and this https://stackoverflow.com/a/8247497 works

ngzhian:
Fixes #1186

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -818,6 +818,7 @@ checkSingleQuotedVariables params t@(T_SingleQuoted id s) =
                 ,"docker" -- like above
                 ,"dpkg-query"
                 ,"jq"  -- could also check that user provides --arg
+                ,"unset"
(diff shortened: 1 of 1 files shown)


koalaman:
Thanks!

PR #1221, opened 2018-05-13
Assignments are okay in SC2094 (fixes #1192)

### LINKED ISSUE #1192: SC2094 complains when no pipes or writes are done.
#### For bugs
- Rule Id: SC2094

```sh

#!/bin/sh
line="" found_fn=""
for fname in "$@"; do
    # shellcheck disable=2162
    while read line; do
        case "$line" in
            foo) found_fn="$fname";;
        esac
    done <"$fname"
done
echo "$found_fn"

```

    $ shellcheck myscript
     
    Line 7:
                foo) found_fn="$fname";;
                              ^-- SC2094: Make sure not to read and write the same file in the same pipeline.
     
    Line 9:
        done <"$fname"
              ^-- SC2094: Make sure not to read and write the same file in the same pipeline.

No errors, there are no pipelines here, and no writing to the file `$fname`

ngzhian:


koalaman:
Nice, thanks! This could theoretically snag on something like `var=foo > $file` but that's not nearly as useful a thing to do.

PR #1222, opened 2018-05-13
Suppress SC2016 for git filter-branch (fixes #196)

### LINKED ISSUE #196: SC2016 should be suppressed for git filter-branch
`git filter-branch` takes a shell snippet to be executed in the context of each git commit.

Currently shellcheck returns SC2016 false positives for variables like `$GIT_COMMIT` in said snippet which are supposed to remain un-expanded.

ngzhian:


koalaman:
Nice, thanks!

PR #1223, opened 2018-05-14
Handle offset references of the form [foo]:bar (fixes #1124)

### LINKED ISSUE #1124: False positive for SC2034 when var used in array parameter expansion as offset or length
- SC2034: 
- My shellcheck version ("online"): 
- [x ] I tried on shellcheck.net and verified that this is still a problem on the latest commit

```sh
#!/bin/bash
arr=(1 2 3 4)
num=2
echo "${arr[@]:0:num}"
echo "${arr[@]:num:1}"
echo "${arr[@]:num}"
```

```
Line 3:
num=2
^-- SC2034: num appears unused. Verify it or export it.
```

ngzhian:


koalaman:
Nice, thanks! At some point, ShellCheck should start parsing parameter expansions properly, instead of having a bunch of string matching logic whenever they're used.

PR #1232, opened 2018-05-23
Add T_SourceCommand to wrap source commands and sourced code

### LINKED ISSUE #1181: the -x option disables SC2039 but it should not
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2039
- My shellcheck version (`shellcheck --version` or "online"): 0.4.6

```sh
$ head foo.sh saywhat.sh 
==> foo.sh <==
#!/bin/sh
source ./saywhat.sh

==> saywhat.sh <==
echo output from saywhat.sh
$ bash foo.sh
output from saywhat.sh
$ dash foo.sh
foo.sh: 2: foo.sh: source: not found
```

```
$ shellcheck -x foo.sh
$ shellcheck foo.sh

In foo.sh line 2:
source ./saywhat.sh
^-- SC2039: In POSIX sh, 'source' in place of '.' is undefined.
^-- SC1091: Not following: ./saywhat.sh was not specified as input (see shellcheck -x).
```

```
$ shellcheck -x foo.sh

In foo.sh line 2:
source ./saywhat.sh
^-- SC2039: In POSIX sh, 'source' in place of '.' is undefined.
$ shellcheck foo.sh

In foo.sh line 2:
source ./saywhat.sh
^-- SC2039: In POSIX sh, 'source' in place of '.' is undefined.
^-- SC1091: Not following: ./saywhat.sh was not specified as input (see shellcheck -x).
```

ngzhian:
Fixes #1181 
Id is not used in T_Include, but I figured i'll leave it there so functions like getId don't have to change.

### DIFF
--- a/src/ShellCheck/AST.hs
+++ b/src/ShellCheck/AST.hs
@@ -134,7 +134,8 @@ data Token =
     | T_Pipe Id String
     | T_CoProc Id (Maybe String) Token
     | T_CoProcBody Id Token
-    | T_Include Id Token Token -- . & source: SimpleCommand T_Script
+    | T_Include Id Token
+    | T_SourceCommand Id Token Token
     deriving (Show)
 
 data Annotation =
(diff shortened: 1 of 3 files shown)


koalaman:
Thanks! I've added a unit test in 467dfe07b, and also switched to two separate IDs just in case some of the parent finding logic gets confused by an ID having itself as its parent ID.

PR #1237, opened 2018-05-25
Check if builtin cd is called

ngzhian:
This adds an additional check, if name of command is `builtin`, we check if the first "flag" is one of the unchecked commands we are about.

### DIFF
--- a/src/ShellCheck/ASTLib.hs
+++ b/src/ShellCheck/ASTLib.hs
@@ -261,7 +261,7 @@ getCommand t =
 getCommandName t = do
     (T_SimpleCommand _ _ (w:rest)) <- getCommand t
     s <- getLiteralString w
-    if "busybox" `isSuffixOf` s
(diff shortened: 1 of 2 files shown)


koalaman:
How about fixing this more generally in `getCommandName`? It already has precedence for doing this for `busybox`:

```
-- Maybe get the command name of a token representing a command
getCommandName t = do
    (T_SimpleCommand _ _ (w:rest)) <- getCommand t
    s <- getLiteralString w
    if "busybox" `isSuffixOf` s
        then
            case rest of
                (applet:_) -> getLiteralString applet
                _ -> return s
        else
            return s
```

This might fix a few other similar cases while hopefully not having serious side effects

ngzhian:
Can't believe i missed that :) that's a much better idea.

ngzhian:
I'm not sure if `isSuffixOf` should be used or `=`, the comparison for busybox uses `isSuffixOf`.

koalaman:
Thanks!

PR #1239, opened 2018-05-27
End column

ngzhian:
I saw your comment [0], I already did a bit of work, figured I might as well upload and see what you think about it :)

[0] https://github.com/koalaman/shellcheck/pull/695#issuecomment-392307999

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -1097,6 +1097,7 @@ readSingleQuoted = called "single quoted string" $ do
             when ('\n' `elem` string && not ("\n" `isPrefixOf` string)) $
                 suggestForgotClosingQuote startPos endPos "single quoted string"
 
+    endPosOfStartId id
     return (T_SingleQuoted id string)
 
 readSingleQuotedLiteral = do
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -1562,6 +1562,7 @@ readDollarVariable = do
     let singleCharred p = do
         n <- p
         value <- wrap [n]
+        endPosOfStartId id
         return $ (T_DollarBraced id value)
 
     let positional = do
@@ -1575,6 +1576,7 @@ readDollarVariable = do
     let regular = do
         name <- readVariableName
         value <- wrap name
+        endPosOfStartId id
         return (T_DollarBraced id value) `attempting` do
             lookAhead $ char '['
             parseNoteAt pos ErrorC 1087 "Use braces when expanding arrays, e.g. ${array[idx]} (or ${var}[.. to quiet)."
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -185,8 +185,9 @@ getNextIdAt sourcepos = do
 
 getNextId :: Monad m => SCParser m Id
 getNextId = do
-    pos <- getPosition
-    getNextIdAt pos
+    start <- startSpan
+    id <- endSpan start
+    return id
 
 data IncompleteInterval = IncompleteInterval SourcePos
 
@@ -202,11 +203,6 @@ endSpan (IncompleteInterval start) = do
         lastId = id,
         positionMap = newMap
     }
-    return id
-
-zeroWidthSpan = do
-    start <- startSpan
-    id <- endSpan start
     return id
 
 addToHereDocMap id list = do
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -136,6 +136,38 @@ almostSpace =
         char c
         return ' '
 
+withNextId :: Monad m => ParsecT s UserState (SCBase m) (Id -> b) -> ParsecT s UserState (SCBase m) b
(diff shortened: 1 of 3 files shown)


ngzhian:
This builds heavily on eatnumber1@'s original pr #695, addressing some of your comments to merge the end position into the existing map.
What's missing is some sort of unit test to verify that end_column is correctly output, will be great if you can point me to existing tests that does similar checks and I can add them there.

src/ShellCheck/Parser.hs
```diff
@@ -136,6 +136,31 @@ almostSpace =
         char c
         return ' '
 
+withNextId :: Monad m => SCParser m (Id -> b) -> SCParser m b
+withNextId p = do
+    start <- getPosition
+    id <- createId
+    fn <- p
```
koalaman: What about the parsers that skip trailing whitespace?

The second `T_NormalWord` for `echo foo      bar` would here be marked with a span of 10 instead of of 3

ngzhian: I haven't tried or looked very deeply, but we can probably avoid this by calling withNextId inside functions that read whitespace, rather than wrapping them. e.g. in readCmdWord, we can do withNextId readNormalWord; spacing, instead of withNextId readCmdWord. Do you think this will work? (I will get to trying it this weekend :)

koalaman: I think that could lead to a lot of issues in the near term. Given that most of the parsers only use the `id` when returning a value, maybe it should be constructed in two phases and without the higher order function, e.g.:

```
parse = do
    start <- getNextId
    whatever
    span <- endId start
    spacing
    return $ T_Stuff span [...] 
```

This has the flexibility of the manual approach and the safety of typechecking

ngzhian: Good point. I implemented this in a series of commits (trying to make changes small).

e496b41 essentially reverts withNextId
0c459ae adds a function that given a id (representing start pos), updates the position map to set the end pos to the current position (maybe need a better name)
3848788 is the demo using this new function when reading dollar variable

I looked at the output using `shellcheck -f json test.sh`, with
```
$ cat test.sh
#!/bin/bash
echo      $1       $2
```

ngzhian: Oh and the output of `shellcheck -f json test.sh` is:
```
[
   {
      "line" : 2,
      "file" : "test2.sh",
      "endLine" : 2,
      "column" : 11,
      "endColumn" : 13,
      "level" : "info",
      "message" : "Double quote to prevent globbing and word splitting.",
      "code" : 2086
   },
   {
      "code" : 2086,
      "message" : "Double quote to prevent globbing and word splitting.",
      "level" : "info",
      "column" : 20,
      "endLine" : 2,
      "endColumn" : 22,
      "file" : "test2.sh",
      "line" : 2
   }
]
```
looks fine, endColumn is exclusive

koalaman: It still looks unchecked to me. Doesn't it still compile if you forget to add `endPosOfStartId id`?

ngzhian: Sorry I probably misunderstood it. Referring to your example previously:
```
parse = do
    start <- getNextId
    whatever
    span <- endPosOfStartId start
    spacing
    return $ T_Stuff span [...] 
```
So change `endPosOfStartId` to return an Id? Or do you mean for `span` to be a different type?

koalaman: It would be convenient if `endPosOfStartId` returned an `Id` for compatibility. The main point is that if you accidentally forget to mark the start position or the end position, the code should not compile. 

This could be one example implementation with different names:

```
data IncompleteInterval = IncompleteInterval Position

startSpan :: m IncompleteInterval
startSpan = IncompleteInterval <$> getPosition

endSpan :: IncompleteInterval -> m Id
endSpan (IncompleteInterval start) = do
  end <- getPosition
  id <- createId
  registerStartAndEnd (start, end) id
  return id

readThing = do
  start <- startSpan
  whatever
  id <- endSpan start
  return $ T_Stuff id
```

In this case (at least if `createId` was hidden), you would not be able to return an AST mode without making sure both the start position and the end position is set. If you forget to do either, the compiler will remind you. This ensures that no refactoring or new code causes any nodes to have missing end positions, since this can be hard to notice during testing.

Some kind of `createZeroWidthId` could be useful while refactoring for code that doesn't mark end positions yet.

ngzhian: Understood:
- followed your suggested API to create a startSpan and endSpan
- updated the existing implementations to use this new API
- `getNextId` now creates an empty span
- make the endPos non-optional in the map

koalaman:
Looks like a great starting point, thanks!

There's still a way to go before this is fully incorporated:
* Set end positions everywhere
* Delete the zero-width functions so they won't be used in the future
* Go over all warnings to ensure they have meaningful end positions
* Expose end positions in the tty and gcc output modes, and on shellcheck.net
* Inevitable extras that come up, like how `T_WhileExpression` throws away information about the `while` keyword, so no test is able to highlight just the `while`

I'm hoping to get to the first two this weekend

PR #1240, opened 2018-05-30
Attempted bats @test support

figure0:


koalaman:
The `bats` branch was abandoned due to lack of use and interest. It would be better off being supported through some third party for now.

PR #1253, opened 2018-06-08
Remove unnecessary dot

sblondon:
The dot after the screenshot in README.md is strange so this commit remove it.


koalaman:
Thanks!

PR #1256, opened 2018-06-13
Do not warn on mv -i (fixes #1251)

### LINKED ISSUE #1251: SC2217: mv does read stdin
I have code similar to the following:

```sh
while read -r filename; do
  mv -i $filename . < /dev/tty
done < <(something)
```

I get SC2217: "Redirecting to 'mv', a command that doesn't read stdin."
In this case `mv` does read stdin and I am not aware of any way to get the desired behaviour without redirecting into `mv`.

I think the check should be refined to test if `mv`/`cp`/`rm`/... is used with `-i`.

ngzhian:


src/ShellCheck/Analytics.hs
```diff
@@ -2856,6 +2857,7 @@ checkPipeToNowhere _ t =
         name <- getCommandBasename cmd
         guard $ name `elem` nonReadingCommands
         guard . not $ hasAdditionalConsumers cmd
+        guard . not $ name == "mv" && "i" `elem` (map snd $ getAllFlags cmd)
```
koalaman: This is also an issue with `cp` and `rm`, so maybe ``name `elem` ["cp", "rm", "mv"] && cmd `hasFlag` "i"``

ngzhian: Updated, ptal

koalaman:
Thanks!

PR #1261, opened 2018-06-15
Do not warn find --help (fixes #1188)

### LINKED ISSUE #1188: SC2185 requires explicit path for 'find --help'
#### For bugs
- Rule Id: SC2185
- I tried on shellcheck.net and verified that this is still a problem on the latest commit

```sh
#!/bin/bash
find --help
```

find --help
^-- SC2185: Some finds don't have a default path. Specify '.' explicitly.

I expected that 'find --help' would not result in any errors.

ngzhian:


koalaman:
Thanks!

PR #1266, opened 2018-06-22
Enable SC2153

junjihashimoto:
SC2153 does not work, ,as getBestMatch returns Nothing when a matched variable is not found.
This PR fixes the bug.


koalaman:
Unfortunately that's by design. This change would emit a warning for any use of any inherited environment variable, which has a very high false positive rate.

ShellCheck recognizes the convention that script local variables have lowercase names while environment variables generally are uppercase. That means it'll warn about `$my_var` but not `$MY_VAR` because the latter is assumed to come from the environment.

Issue #1200, opened 2018-05-03
SC2032 and SC2033 false positives when conditionally defining function


tdmalone:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2032/SC2033
- My shellcheck version (`shellcheck --version` or "online"): 0.4.7
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit
- [ ] It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/usr/bin/env bash

if [ "$( sudo -v )" ]; then
  sudo mv terraform /usr/bin/terraform
else
  terraform() { "${TRAVIS_BUILD_DIR}/terraform" "$@"; }
  export -f terraform
fi

```

#### Here's what shellcheck currently says:

```
Line 4:
  sudo mv terraform /usr/bin/terraform
          ^-- SC2033: Shell functions can't be passed to external commands.
 
Line 6:
  terraform() { "${TRAVIS_BUILD_DIR}/terraform" "$@"; }
  ^-- SC2032: Use own script or sh -c '..' to run this from sudo.
```

#### Here's what I wanted or expected to see:

```
No issues detected!
```

tdmalone:
I'm open to the fact that I might be doing something wrong, but I think this is a false positive. I'm conditionally creating the function, so it shouldn't even exist at the point I'm trying to run sudo instead.

I also realise this might be very hard to detect, and I will use a directive for now.

Dmole:
I think if you or some calling script wrap your whole code in brackets it will create the function and export it then check the condition so I think it's a valid warning. You should probably just re-name your function.

PR #1281, opened 2018-07-05
First pass at simple control flow sensitive external function use

ngzhian:
I modified checkFunctionsUsedExternally slightly to do the functions collection and the checking of external usage in the same pass, which is more accurate when it comes to control flow. Prototyped it in this PR, let me know if you think it's an approach that makes sense?

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -1781,6 +1781,44 @@ checkFunctionsUsedExternally params t =
             info definitionId 2032 $
               "Use own script or sh -c '..' to run this from " ++ cmd ++ "."
 
+prop_a1 =
+  verifyTree a "foo() { :; }; sudo foo"
+prop_a2 =
+  verifyTree a "alias f='a'; xargs -n 1 f"
+prop_a3 =
+  verifyNotTree a "f() { :; }; echo f"
+prop_a4 =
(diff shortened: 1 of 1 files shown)


koalaman:
It definitely makes sense, but I'm not sure this is a significant source of false positives for this warning. I wasn't able to find any examples of this happening from my collection of github scraped scripts (not that it's a common warning to begin with)

ngzhian:
I can close this and leave it around as a template for how to implement other checks. From your scripts, which false positive warning will be worth fixing? (Also forgot to mention that this fixes #1200)

koalaman:
I haven't gone through them in a while, I mostly use them to compare ShellCheck output before and after a change to see whether it has unintended consequences. If you want to look through them for false positives, here are 3000+ scripts I've scraped from popular github projects: https://www.vidarholen.net/~vidar/github_scripts.tar.gz

These are a good representation for the kind of production scripts people would shellcheck. I admit I haven't gone through and manually verified the individual licenses, but all the scripts are clearly named after the public repository in which they were found.

I have a similar collection scraped from pastebin links posted on the #bash@freenode IRC channel which do a better job representing newbies' troublesome scripts. I'm happy to use them myself since they were posted in a public forum with no reasonable expectation of privacy, but I don't believe I have any rights to redistribute them.

PR #1282, opened 2018-07-05
Add instructions to install linux binary

kenden:
If a user wants to install a fixed version of `shellcheck` on a linux system, it was not obvious where to find the instructions.
This PR adds some documentation about that in the `README.md` file


koalaman:
Thanks!

PR #1298, opened 2018-07-19
Fix false positive when indexing into array in cond

### LINKED ISSUE #1268: BUG: SC1073 triggered on valid code and shellcheck stops processing
#### For bugs
- Rule Id: SC1073,SC1019,SC1020,SC1072
- My shellcheck version: online

```sh
#!/usr/bin/env bash
declare -A ARRAY=([alpha]=a [bravo]=b)
IDX=--alpha
[[ -v ARRAY[${IDX#--}] ]] && echo "ARRAY[${IDX#--}]=${ARRAY[${IDX#--}]}"

```

```
Line 4	SC1073: Couldn't parse this test expression. Fix to allow more checks.
Line 4	SC1019: Expected this to be an argument to the unary condition.
Line 4	SC1020: You need a space before the ]].
Line 4	SC1072: Missing space before ]. Fix any mentioned problems and try again.
```

Nothing. AFAIK the code is valid and executes with no issues. This is the output with ```xtrace``` enabled:
```
+ ARRAY=([alpha]=a [bravo]=b)
+ declare -A ARRAY
+ IDX=--alpha
+ [[ -v ARRAY[alpha] ]]
+ echo 'ARRAY[alpha]=a'
ARRAY[alpha]=a
```

I think the problem relies on shellcheck not being able to recognize the parameter substitution used on the array subscript of the ```[[ -v ... ]]``` test. If I edit the condition to ```[[ -v ARRAY[$IDX] ]]``` I get the ```No issues detected!``` happy ending :-) This breaks the code, of course.

Unfortunately, this issue is critical for me. Shellcheck refuses to process anything else until I "fix" my code, and no directives disabling the reported conditions allow shellcheck to proceed.

Any suggestions, other than selectively commenting the "offending" lines of code?

ngzhian:
Fixes #1268 

This checks that when the T_NormalWord read in a unary cond ends in a ']' that we did not have a starting '[', which would signify that this is an array index.

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -520,7 +520,7 @@ readConditionContents single =
         notFollowedBy2 (try (spacing >> string "]"))
         x <- readNormalWord
         pos <- getPosition
-        when (endedWith "]" x) $ do
+        when (endedWith "]" x && notArrayIndex x) $ do
             parseProblemAt pos ErrorC 1020 $
                 "You need a space before the " ++ (if single then "]" else "]]") ++ "."
             fail "Missing space before ]"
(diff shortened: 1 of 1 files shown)


koalaman:
It does the job for 95% of `-v name[anything]`  type scenarios. Thanks!

PR #1302, opened 2018-07-23
Fix #949 (failing on @ in function names)

### LINKED ISSUE #909: shellcheck can't parse actual bash function name
#### For bugs
- Rule Id SC1089
- My shellcheck version 0.4.6

```sh
# can we really do funky crap like this?
function [}{*(){
   echo "Let me take you to, funkytown!"
}
[}{*    # why yes, we can!
# though editor auto-indent modes may punish us

```

emits an SC1089 on the trailing '}' of function definition

Oh, I don't know. Maybe something like "bash allows nearly every character (printable or not) as a function name, but I'm just going to quit now if you're going to write that sort of code."

Inspired by seeing actual function names containing '?' in the wild. Documented more here:
http://stackoverflow.com/questions/28114999/what-are-the-rules-for-valid-identifiers-e-g-functions-vars-etc-in-bash/44041384#44041384

### LINKED ISSUE #949: SC1036 error on valid Bash function name
#### For bugs
- Rule Id (if any, e.g. SC1000): SC1036
- My shellcheck version (`shellcheck --version` or "online"): 0.3.7

```sh
#!/usr/bin/env bash
@func() { echo "Hello"; }
@func
```

```
Line 2:
@func() { echo "Hello"; }
     ^-- SC1036: '(' is invalid here. Did you forget to escape it?
     ^-- SC1088: Parsing stopped here. Invalid use of parentheses?
```

No errors.

I've tested this against Bash 3.2.48 and 4.3.48. Both handle the script just fine. I don't believe that the parenthesis can be escaped, nor would I want that because then potentially the parenthesis would be part of the function name. I did try escaping a parenthesis just for fun and that produced different errors and didn't work in Bash.

pjeby:
'@' was previously mentioned in 895ba3133783e23e3031f1991cfa63c154c79dce as a character needed to fix #909, but was not included in the actual change at that time.

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -66,7 +66,7 @@ doubleQuote = char '"'
 variableStart = upper <|> lower <|> oneOf "_"
 variableChars = upper <|> lower <|> digit <|> oneOf "_"
 -- Chars to allow in function names
-functionChars = variableChars <|> oneOf ":+?-./^"
+functionChars = variableChars <|> oneOf ":+?-./^@"
 -- Chars to allow in functions using the 'function' keyword
 extendedFunctionChars = functionChars <|> oneOf "[]*=!"
 specialVariable = oneOf "@*#?-$!"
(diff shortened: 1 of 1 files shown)


koalaman:
Seems reasonable and does not appear to interfere with anything. Thanks!

PR #1304, opened 2018-07-24
replace CS2006 message 

p-alik:
my colleague @andk discovered the issue. I just provide his proposal.

- test script `t.sh`
```bash
#!/usr/bin/env bash
DATE=`date`
echo "$DATE"

```

- running test without PR
```bash
$ stack exec -- shellcheck t.sh

In t.sh line 2:
DATE=`date`
     ^----^ SC2006: Use $(..) instead of legacy `..`.
```

PR uses the headline of https://github.com/koalaman/shellcheck/wiki/Sc2006

- running test with PR
```bash
$ stack exec -- shellcheck t.sh

In t.sh line 2:
DATE=`date`
     ^----^ SC2006: Use $(STATEMENT) instead of legacy `STATEMENT`.

```


koalaman:
According to the wiki log, someone replaced `..` with `STATEMENT` in the wiki because they felt it was easily confused with "parent dir".

There is some validity to that, but I'm not a fan of the edit. I've instead changed the wiki to use `...` which mimics usage in POSIX, and updated the emitted warning to match.

Thanks for pointing out the discrepancy!

PR #1311, opened 2018-08-06
Use regex to match special flags for printf

### LINKED ISSUE #1310: Regression SC2183 when printf is given '*' arguments in the format string
#### For bugs
- Rule Id (if any, e.g. SC1000): **SC2183**
- My shellcheck version (`shellcheck --version` or "online"): **0.5.0**
- [**yes**] I tried on shellcheck.net and verified that this is still a problem on the latest commit
- [**n/a**] It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue

```sh

#!/usr/bin/env bash
printf "%-*s  %s\\n" "$max_arg" "$first" "$line"

```

```
In functions/common_logging.sh line 166:
      printf "%-*s  %s\\n" "$max_arg" "$first" "$line"
             ^-- SC2183: This format string has 2 variables, but is passed 3 arguments.
```

I expect there to be no error as this is valid. This is a regression of #1184 , also version 0.4.7 does not flag this line as expected.

ngzhian:
Fixes #1310


koalaman:
Nice, thanks!

PR #1314, opened 2018-08-11
Make data in Interface more opaque

ngzhian:
Create a new data type Fix that represents suggested fix for a warning. Right now there is only one variant, which is a Surround String, representing a fix that surrounds the warning with a string.
This is created using `makeCommentWithFix`

Also worked on making the data types in interface more opaque, exposing all the record accessor functions. Not sure if this is alright, please take a look?

### DIFF
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -73,7 +73,7 @@ data Options = Options {
 defaultOptions = Options {
     checkSpec = emptyCheckSpec,
     externalSources = False,
-    formatterOptions = FormatterOptions {
+    formatterOptions = newFormatterOptions {
         foColorOption = ColorAuto
     }
 }
--- a/src/ShellCheck/Analyzer.hs
+++ b/src/ShellCheck/Analyzer.hs
@@ -30,7 +30,7 @@ import qualified ShellCheck.Checks.ShellSupport
 
 -- TODO: Clean up the cruft this is layered on
 analyzeScript :: AnalysisSpec -> AnalysisResult
-analyzeScript spec = AnalysisResult {
+analyzeScript spec = newAnalysisResult {
     arComments =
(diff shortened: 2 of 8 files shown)


koalaman:
The opaqueness is exactly what I had in mind.

I'm not sure if the Fix is generally applicable enough though. Which other fixes will involve surrounding a string with something? What if you wanted to fix `$foo/bar` into `"$foo/bar"`?

matthewpersico:
Strictly speaking, shellcheck currently only needs "$foo"/bar to fix that kind of warning currently. Therefore, I say do not try and get fancy - only quote what's needed and do not try to figure out what's wanted; THAT would be a rabbit hole from which you would never return.

ngzhian:
I removed the Fix part, this PR is only only about making the data types more opaque.
I agree with the "do no try and get fancy part", start with simple fixes that we can actually suggest. And I also think just having surround is insufficient, I'll work on another pr that probably adds a more general, Replace start end string type.

matthewpersico:
This is very exciting work. I currently have an emacs function that, when I type 'd' on a shellcheck error in a flycheck emacs buffer, inserts that appropriate disable comment above the statement in question. When you get around to the actual fixing of items, I'd like to see how it's done so I can integrate it into emacs. Maybe we can talk offline - I am at matthew.persico@gmail.com.

koalaman:
Merged, thanks!

koalaman:
I also agree. But obviously `diff/patch` would not be simplified by making it output and interpret a series of human-relatable actions:

    $ diff file1 file2
    Add quotes around the variables on line 42
    Remove the unquoted commas in the assignment to myarray
    Replace the unicode quotes with ascii quotes on line 64

It's way more simple to have it work by replacing arbitrary text spans.

Issue #1272, opened 2018-06-27
Confused about arrays


matthewpersico:
#### For bugs
- Rule Id (if any, e.g. SC1000): 
- My shellcheck version (`shellcheck --version` or 'online'): online
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh
#!bash

edit='emacsclient -c'
declare -a editcmd
if [ -n "$edit" ]
then
    read -r -a editcmd < <(echo "$edit")
fi

if [[ "$editcmd" =~ emacs\  ]]
then
    editcmd="$editcmd -g +100+100"
fi

if [ -n "$editcmd" ]
then
    "$editcmd" &
fi

```

#### Here's what shellcheck currently says:
```

```
#### Here's what I wanted or expected to see:

I expected it to complain about using "$editcmd" instead of "${editcmd[@]}" in all the places where the former was used. Am I missing something?

kerframil:
I think that this ought to trigger SC2128. Consider the following sample.

```sh
#!/bin/bash
var=(foo bar baz)
[[ $var == foo\ * ]]
```

Shellcheck reports on it as follows:-

```
[[ $var == foo\ * ]]
   ^-- SC2128: Expanding an array without an index only gives the first element.
```

The use of read -ra seems to trip shellcheck up.

kerframil:
See also #679.

koalaman:
Fixed by @ngzhian. Thanks!

PR #1324, opened 2018-08-27
Understand array variable declaration in read (fixes #679)

### LINKED ISSUE #679: SC2154 doesn't understand read -a array [more options]
``` bash
read -a foo -r <<<"foo bar"
read -r -a bar <<<"foo bar"
read -a baz -r baz <<<"foo bar"
echo "$foo $bar $baz"
# Line 4:
# echo "$foo $bar $baz"
#       ^-- SC2154: foo is referenced but not assigned.
```

I'm guessing fixing this bug will make shellcheck identify them as arrays, so no point in creating issues about missing array warnings?

ngzhian:
It used to only treat all trailing variables in read as varaible
declarings, but an array variable can be declared in other positions:

    read -a foo -r

foo is a declared variable, and multiple such variables can be declared.

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -709,6 +709,7 @@ prop_checkArrayWithoutIndex5 = verifyTree checkArrayWithoutIndex "a[0]=foo; echo
 prop_checkArrayWithoutIndex6 = verifyTree checkArrayWithoutIndex "echo $PIPESTATUS"
 prop_checkArrayWithoutIndex7 = verifyTree checkArrayWithoutIndex "a=(a b); a+=c"
 prop_checkArrayWithoutIndex8 = verifyTree checkArrayWithoutIndex "declare -a foo; foo=bar;"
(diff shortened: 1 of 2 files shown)


koalaman:
Nice! This is one of the uglier parts of ShellCheck. The function returns a tuple consisting of four items describing an assignment. Given e.g.

    declare foo=bar

it will return

    (
      BaseCommand :: Token,     -- The command/structure assigning the variable, i.e. declare foo=bar
      AssignmentToken :: Token, -- The specific part that assigns this variable, i.e. foo=bar
      VariableName :: String,   -- The variable name, i.e. foo
      VariableValue :: DataType -- A description of the value being assigned, i.e. "Literal string with value foo"
    )

In the case of `read -a foo`, you can have it specify a value of `DataArray SourceExternal` to indicate "an array with data from some external source". This way, ShellCheck will know that this is an array variable, thereby fixing #1272 as well.

ngzhian:
Updated, take a look whenever you have time :)

koalaman:
Perfect, thanks!

PR #1332, opened 2018-09-07
cleanup

leogtzr:
some cleanup :)

### DIFF
--- a/nextnumber
+++ b/nextnumber
@@ -1,13 +1,18 @@
 #!/usr/bin/env bash
 # TODO: Find a less trashy way to get the next available error code
-if ! shopt -s globstar
-then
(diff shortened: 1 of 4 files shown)


koalaman:
There are some improvements here, like quoting string literals and using a local variable for `data`

However, this PR consistently replaces portable short options with non-portable long options. It's true that the use of e.g. `grep -o` restricted portability to begin with, but I would rather go the other direction of increased portability. 

Several of the other changes, like switching to K&R indenting style, consistently using curly braces on variables, and adding explicit exits, are debatable stylistic issues. The current code is an indication of which style I personally prefer.

leogtzr:
Alright, I will change back some of the things you mentioned and send the PR back to you :) thanks

PR #1335, opened 2018-09-10
Doctest new build

phadej:
This PR started with a need to make ShellCheck compilation a little faster.

I changed tests to be run through `doctest` http://hackage.haskell.org/package/doctest, therefore they are not compiled in anymore. That removed the need for `striptests` and reduced compilation time from *140sec* to *90sec* on my machine.

---

The 3rd and 4th commits are rework of `Dockerfile` to use GHC-8.4.3 and cabal new-build (https://cabal.readthedocs.io/en/latest/nix-local-build-overview.html). One reason is to try whether `cabal new-build` can handle this workflow, as it will become the default in the near future.

Another benefit is that we can patch `aeson` and `regex-tdfa` quite simply when using `new-build`. This both reduces full (docker) build time, as well as the size of final library (as -O2 makes binaries quite large). On my machine the final binary size is 5937096 Mb.

---

Second to last commit fixes `quicktest`

---

Finally there's small improvement in `myPreSDist`. However, I'd rather remove that hook, and perform `pandoc` "manually" in the packaging script.


koalaman:
It sounds great to not have to deal with dummy test names while also removing any reliance on TH!

How does the workflow have to change due to this? I see for example that `./quicktest` used to run the tests regardless of compilation status but now does nothing, and `cabal install --dependencies-only && cabal build`  and `cabal test` both fail with `doctest >=0.16.0 && <0.17, template-haskell -any`

phadej:
@koalaman, I could write a small howto develop section if I can assume some versions of `cabal install` (Ubuntu 18.04 has `cabal-install-1.24`).

- yes `quicktest` has a regression that you have to build project successfully *once*. This is a trade-off.

- `cabal install --dependencies-only && cabal build` fails: how about `cabal install --dependencies-only --enable-tests && cabal build` ?

koalaman:
That did indeed work. It looks good to me. I just want to test it on the major distros to ensure good compatibility, and I'll have that done as soon as I can. Don't worry about the merge conflicts, I'll fix that.

koalaman:
The tests via doctest in this build takes over 5 minutes to run, whereas the previous tests took less than 30 seconds on a less powerful machine.

As far as I've been able to tell, a major factor is that the properties are run 100 times as per QuickCheck, whereas the existing build only runs them once since they do not depend on additional input.

I haven't been able to find much documentation for this package, but I filed an issue https://github.com/sol/doctest/issues/212 to see if there's any way to get around this.

koalaman:
Looks like I'm just referencing some old bug from four years ago, iteration count should not be the issue.

Are there any other ways to bring this time down? 5+ minutes is far too much.

phadej:
@koalaman I see. For `doctest` approach is unfortunately slower, though both old and new `quicktest` essentially load ShellCheck into GHCi and run the tests.

koalaman:
That's too bad, this kind of setup would be awesome. I could live with tests taking twice as long, but 10x is too much.

phadej:
@koalaman I made a small change, not test runs considerably faster:

```
  24,48s user 0,67s system 101% cpu 24,780 total
```

you can even change "predicates" to bail with something more informative than just `FAIL`.

koalaman:
I merged this into a separate branch [doctest](https://github.com/koalaman/shellcheck/tree/doctest) and it's looking great, except under Docker it's failing with a rather opaque message:

```
[...]
ShellCheck.Regex
Paths_ShellCheck
doctests: fd:14: hGetLine: end of file
Test suite doctests: FAIL
Test suite logged to: dist/test/ShellCheck-0.6.0-doctests.log
0 of 1 test suites (0 of 1 test cases) passed.
test failed
```

The command  used was:

    sudo docker run -v "$PWD:/mnt" -it debian:stable sh -c 'cd /mnt && apt-get update && apt-get install -y cabal-install && test/buildtest'

Any idea how to debug this? The `doctest` binary does not appear to have any flags.

austin987:
Try calling `test/buildtest` with `sh -x test/buildtest`?

koalaman:
The problem occurs regardless of how the `doctests` binary is invoked.

phadej:
Sorry for not replying.

```
Test suite logged to: dist/test/ShellCheck-0.6.0-doctests.log
```

probably contains the reason.

I'm however don't have time to push this forward.

PR #1357, opened 2018-10-08
Remove 'du' from list of non-reading commands

Jellyfrog:
Remove 'du' from list of non-reading commands due to it being able to read from stdin:

>  --files0-from=F
> summarize disk usage of the NUL-terminated file names specified in file F; if F is -, then read names from standard input

### DIFF
--- a/src/ShellCheck/Data.hs
+++ b/src/ShellCheck/Data.hs
@@ -79,7 +79,7 @@ commonCommands = [
 
 nonReadingCommands = [
     "alias", "basename", "bg", "cal", "cd", "chgrp", "chmod", "chown",
-    "cp", "du", "echo", "export", "false", "fg", "fuser", "getconf",
+    "cp", "echo", "export", "false", "fg", "fuser", "getconf",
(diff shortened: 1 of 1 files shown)


Jellyfrog:
@koalaman maybe there should be a separate list with arguments required to "activate" stdin-mode.
For example an array like `["du" = "--files0-from=-", ...]`

koalaman:
Yes, there should probably be something like that. Currently, it's slightly ad-hoc in the function itself:

    https://github.com/koalaman/shellcheck/blob/master/src/ShellCheck/Analytics.hs#L2886

Given that  `--files0-from`  is a relatively limited use case, it would probably better to do something similar to that to still allow errors in the more common case of `... | du`

zertrin:
For information, even if the man page doesn't say it explicitly, it is also possible to pass stdin to the `--exclude-from=-` option (obviously the `--files0-from=-` should not be used at the same time).

koalaman:
I have added a command specific suppression in 3e50a2fce8. Thanks!

PR #1361, opened 2018-10-14
Prototype fix

ngzhian:
Here's a prototype for Fix that is ready for suggestions. The API is pretty simple:
- a Fix is a list of Replacement
- a Replacement has a start Position, end Position (exclusive), and a String to replace the this range with

I added the fix to the json output quite simply, (since it doesn't apply the fix, only specifies what the fix is).

The work on ed-style output, applying fixes, checking for overlaps, can probably come in another PR.

With this test file:
```
#!/bin/bash
DEFS=( -DBADVPN_THREAD_SAFE=0 -DBADVPN_LINUX -DBADVPN_BREACTOR_BADVPN -D_GNU_SOURCE )
echo "${DEFS[1]}"

CC=gcc
CFLAGS="cat a"
"${CC}" -c ${CFLAGS}

cd "$DIR"

CUR_HASH=`$pm2 prettylist | grep "revision" | cut -d: -f2 | tr -d " ,'"``"'"`

echo "$CUR_HASH"
```
The json output is like so:
```
[
   {
      "replaceWith" : [
         {
            "endColumn" : 8,
            "replaceWith" : "\"",
            "line" : 2,
            "endLine" : 2,
            "column" : 8
         },
         {
            "column" : 31,
            "endLine" : 2,
            "line" : 2,
            "replaceWith" : "\"",
            "endColumn" : 31
         }
      ],
      "file" : "test.sh",
      "message" : "The = here is literal. To assign by index, use ( [index]=value ) with no spaces. To keep as literal, quote it.",
      "endColumn" : 30,
      "column" : 8,
      "level" : "warning",
      "endLine" : 2,
      "line" : 2,
      "code" : 2191
   },
   {
      "code" : 2086,
      "line" : 7,
      "endLine" : 7,
      "level" : "info",
      "column" : 12,
      "endColumn" : 21,
      "file" : "test.sh",
      "message" : "Double quote to prevent globbing and word splitting.",
      "replaceWith" : [
         {
            "endColumn" : 12,
            "replaceWith" : "\"",
            "line" : 7,
            "endLine" : 7,
            "column" : 12
         },
         {
            "column" : 22,
            "endLine" : 7,
            "line" : 7,
            "replaceWith" : "\"",
            "endColumn" : 22
         }
      ]
   },
   {
      "endColumn" : 10,
      "message" : "Use 'cd ... || exit' or 'cd ... || return' in case cd fails.",
      "file" : "test.sh",
      "replaceWith" : [
         {
            "endColumn" : 11,
            "line" : 9,
            "replaceWith" : " || exit",
            "endLine" : 9,
            "column" : 11
         }
      ],
      "endLine" : 9,
      "level" : "warning",
      "column" : 1,
      "line" : 9,
      "code" : 2164
   },
   {
      "endLine" : 11,
      "column" : 10,
      "level" : "style",
      "endColumn" : 73,
      "replaceWith" : [
         {
            "replaceWith" : "$(",
            "line" : 11,
            "endColumn" : 11,
            "column" : 10,
            "endLine" : 11
         },
         {
            "column" : 73,
            "endLine" : 11,
            "line" : 11,
            "replaceWith" : ")",
            "endColumn" : 74
         }
      ],
      "message" : "Use $(...) notation instead of legacy backticked `...`.",
      "file" : "test.sh",
      "code" : 2006,
      "line" : 11
   },
   {
      "level" : "warning",
      "column" : 11,
      "endLine" : 11,
      "message" : "pm2 is referenced but not assigned.",
      "file" : "test.sh",
      "replaceWith" : null,
      "endColumn" : 15,
      "code" : 2154,
      "line" : 11
   },
   {
      "endColumn" : 78,
      "replaceWith" : [
         {
            "endLine" : 11,
            "column" : 73,
            "endColumn" : 74,
            "replaceWith" : "$(",
            "line" : 11
         },
         {
            "column" : 78,
            "endLine" : 11,
            "line" : 11,
            "replaceWith" : ")",
            "endColumn" : 79
         }
      ],
      "file" : "test.sh",
      "message" : "Use $(...) notation instead of legacy backticked `...`.",
      "endLine" : 11,
      "column" : 73,
      "level" : "style",
      "line" : 11,
      "code" : 2006
   }
]
```

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -241,6 +241,31 @@ isCondition (child:parent:rest) =
             T_UntilExpression id c l -> take 1 . reverse $ c
             _ -> []
 
+-- helpers to build replacements
+replace_start id params n r =
+    let tp = tokenPositions params
+        (start, _) = tp Map.! id
+        new_end = start {
+            posColumn = posColumn start + n
+        }
+    in
+    [R start new_end r]
+replace_end id params n r =
+    -- because of the way we count columns 1-based
+    -- we have to offset end columns by 1
+    let tp = tokenPositions params
+        (_, end) = tp Map.! id
+        new_start = end {
+            posColumn = posColumn end - n + 1
+        }
+        new_end = end {
+            posColumn = posColumn end + 1
+        }
+    in
+    [R new_start new_end r]
+surround_with id params s =
+    (replace_start id params 0 s) ++ (replace_end id params 0 s)
+
 prop_checkEchoWc3 = verify checkEchoWc "n=$(echo $foo | wc -c)"
 checkEchoWc _ (T_Pipeline id _ [a, b]) =
     when (acmd == ["echo", "${VAR}"]) $
@@ -1334,10 +1359,10 @@ checkPS1Assignments _ _ = return ()
 prop_checkBackticks1 = verify checkBackticks "echo `foo`"
 prop_checkBackticks2 = verifyNot checkBackticks "echo $(foo)"
 prop_checkBackticks3 = verifyNot checkBackticks "echo `#inlined comment` foo"
-checkBackticks _ (T_Backticked id list) | not (null list) =
+checkBackticks params (T_Backticked id list) | not (null list) =
     addComment $
         makeCommentWithFix StyleC id 2006  "Use $(...) notation instead of legacy backticked `...`."
-            ((replaceStart 1 "$(") ++ (replaceEnd 1 ")"))
+            ((replace_start id params 1 "$(") ++ (replace_end id params 1 ")"))
     -- style id 2006 "Use $(...) notation instead of legacy backticked `...`."
 checkBackticks _ _ = return ()
 
@@ -1643,7 +1668,7 @@ checkSpacefulness params t =
                     "This default assignment may cause DoS due to globbing. Quote it."
             else
                 makeCommentWithFix InfoC (getId token) 2086
-                    "Double quote to prevent globbing and word splitting." (surroundWith "\"")
+                    "Double quote to prevent globbing and word splitting." (surround_with (getId token) params "\"")
                 -- makeComment InfoC (getId token) 2086
                 --     "Double quote to prevent globbing and word splitting."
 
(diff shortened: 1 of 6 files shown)


Superpat:
It would be nice to get ed style output instead of modifying the base file directly.

ngzhian:
Can you give an example of "ed style output", is it something like "s/foo/bar/g"?
In this pr the fixed source is printed to stdout without updating the file in place (was just hacking to see it working :))

koalaman:
This is general enough to output ed or diff syntax, it doesn't modify the file by itself.

I don't think all modifications can be directly related to the comment's single token. You can't update two related nodes at once for example.

Let's expose the token position map in that `params` parameter and let the concept of "N characters from start/end" just be convenience functions that take the token and N, and returns the appropriate range. This also trivially extends to e.g. `replace Token String :: Fix` or `delete Token`

This would also make it easier to generate a Fix from the parser, which mostly works on positions

Superpat:
On 10/18, Ng Zhi An wrote:
> Can you give an example of "ed style output", is it something like "s/foo/bar/g"?
Sort of, but since the diff program knows exactly what changed it's
more precise. A perfect example would be the output of the diff(1)
or git-diff(1) programs. This kind of output can be use with patch(1)
or ed(1) to update a program. It also comes in handy to send patches
by email, like git-format-patch(1) / git-send-email(1) do.

> In this pr the fixed source is printed to stdout without updating the file in place (was just hacking to see it working :))
That would probably be fine, since you can run `diff script.sh
checked_script.sh` and get ed(1) formatted output.

API wise, I'd like to see something like `shellcheck --format ed script.sh` 
or `shellcheck --format patch script.sh`, since it's really simply 
another way of explaining the errors, just a machine 
(and human) readable version.

> https://github.com/koalaman/shellcheck/pull/1361#issuecomment-431246127
--
Patrick Marchand
https://www.patrick.marchand.ski

koalaman:
This is awesome! I merged this as a new branch named [autofix](https://github.com/koalaman/shellcheck/tree/autofix)

There are some things left to do, and I patched up a couple temporarily
* Lines with tabs don't fix correctly
* All the patch application code should probably be in a separate module
* Lines with fixes that span multiple lines are wonky (I filtered them out)
* Lines with multiple fixes need better handling (I made it only show the first)
* Memory usage ballooned (I added a relatively heavy dependency on deepseq)

It looks fantastic when it outputs useful replacements, it won't take a ton more to be useful

PR #1396, opened 2018-11-21
More warnings for unquoted unary tests

### LINKED ISSUE #750: SC2070 (test -n without quoting) needs to be generalized
SC2070 detects problems of this type: [ -n $var ] which should be changed to [ -n "$var" ], because the former will always be true, even if var is empty, or only contains whitespace.
See https://github.com/koalaman/shellcheck/wiki/SC2070 for full rationale.

However, shellcheck only complains about -n and [ ], , while the problem is more general and extra checks should be added.

Firstly, [ ] is the same as 'test', and 'test -n $var' is equally wrong.

Secondly, there are other unary checks in 'test' that exhibit problems when the tested variable is unquoted and is empty or whitespace-only, for example -f (file), -d (directory), ...
In each of these cases, an unquoted variable will result in 'true' if the variable is empty or whitespace-only.

### LINKED ISSUE #1353: Stronger warning on test -z

```sh
#!/bin/bash 
somestring=$(ls /data)

if [ -z $somestring ]; then
  echo somestring is empty
else
  rm -rf /data/$somestring
fi
```

```
Line 4	SC2086: Double quote to prevent globbing and word splitting.
Line 7	SC2086: Double quote to prevent globbing and word splitting.
```

Because the shell will elide `$somestring` if undefined or empty, the `if` command is rendered
`if '[' '-z' ']'; then`
creating the 1-arg version of test, with `'-z'` being the string under test, which then returns success (0).

It *appears* superficially that the code is working, because empty/missing strings pass the test, but not in the intended sense of test evaluating the length of the second arg.

Additionally, if `$somestring` contains whitespace, test will fail with syntax error ("binary operator expected" or "extra argument" from GNU coreutils test, "too many arguments" from bash).

#### Proposal:

shellcheck already identifies SC2086 and possibly SC2154 for the -z line. Please add a new warning or error specifically to handle test -z with unquoted variable expansion.

ngzhian:
Fixes #1353
Fixes #750

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -876,8 +876,8 @@ prop_checkUnquotedN2 = verify checkUnquotedN "[ -n $cow ]"
 prop_checkUnquotedN3 = verifyNot checkUnquotedN "[[ -n $foo ]] && echo cow"
 prop_checkUnquotedN4 = verify checkUnquotedN "[ -n $cow -o -t 1 ]"
 prop_checkUnquotedN5 = verifyNot checkUnquotedN "[ -n \"$@\" ]"
-checkUnquotedN _ (TC_Unary _ SingleBracket "-n" (T_NormalWord id [t])) | willSplit t =
-       err id 2070 "-n doesn't work with unquoted arguments. Quote or use [[ ]]."
+checkUnquotedN _ (TC_Unary _ SingleBracket op (T_NormalWord id [t])) | willSplit t && op `elem` unaryTestOps =
+       err id 2070 (op ++ " doesn't work with unquoted arguments. Quote or use [[ ]].")
 checkUnquotedN _ _ = return ()
 
 prop_checkNumberComparisons1 = verify checkNumberComparisons "[[ $foo < 3 ]]"


koalaman:
You currently get quite a lot of these kinds of things:

```
        if [ ! -d ${LOGSDIR} ] ; then
                  ^--------^ SC2070: -d doesn't work with unquoted arguments. Quote or use [[ ]].
                  ^--------^ SC2086: Double quote to prevent globbing and word splitting.
```
 
```
 if [ -f $TMPFILE ]; then rm -f $TMPFILE; fi
          ^------^ SC2070: -f doesn't work with unquoted arguments. Quote or use [[ ]].
          ^------^ SC2086: Double quote to prevent globbing and word splitting.
```

We should probably disable SC2086 if there's a 2070 mirroring the same advice

koalaman:
I think it would be better to check at SC2086 and then not output it (or alternatively, have that be where SC2070 is output).

There's currently no other logic that tries to resolve interactions between multiple warnings after they've been emitted, and it would be nice to keep it that way.

iamlazytom:
tried this package and at my level very helpful. as a bonus set me on a path to find a solution and ended up signing up to GitHub. Problem I had was shellcheck sc2039 and trying to resolve that found that echo is allowed to do what printf cannot. Also found 'read' may require '-'r' but you can only use read -r or -p or -rp . 
Hope this file shows problems that can be solved for printf. analyzing # shellcheck should be interesting. 
[resultnovariables.txt](https://github.com/koalaman/shellcheck/files/4496177/resultnovariables.txt)

koalaman:
I'll leave this for SC2086. Thanks!

PR #1397, opened 2018-11-22
Disable smart typography extension for markdown input

### LINKED ISSUE #1392: En dash in front of long options in man page
#### For bugs
- My shellcheck version: 0.5.0

```sh
$ man shellcheck | grep '–'
       -a, –check-sourced
       -C[WHEN], –color[=WHEN]
              For TTY output, enable colors always, never or auto.  The default is auto.  –color without an argument
              is equivalent to –color=always.
       -e CODE1[,CODE2...], –exclude=CODE1[,CODE2...]
       -f FORMAT, –format=FORMAT
       -s shell, –shell=shell
       -V, –version
       -x, –external-sources
```

The long options shown above are prefixed with `–` ([en dash](https://www.fileformat.info/info/unicode/char/2013/index.htm)). If I copy one of those options verbatim it's treated as a file:

```sh
$ shellcheck –check-sourced *.sh 
–check-sourced: –check-sourced: openBinaryFile: does not exist (No such file or directory)
```

All the long options should be prefixed with `--` (two hyphens).

ngzhian:
Fixes #1392

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -33,4 +33,4 @@ myPreSDist _ _ = do
   putStrLn $ "pandoc exited with " ++ show result
   return emptyHookedBuildInfo
   where
-    pandoc_cmd = "pandoc -s -t man shellcheck.1.md -o shellcheck.1"
+    pandoc_cmd = "pandoc -s -f markdown-smart -t man shellcheck.1.md -o shellcheck.1"


koalaman:
Are all these extra files intended to be there?

ngzhian:
Good call, they are not supposed to be there, I'll on holiday right now and
will get to this when I'm back!

koalaman:
Take your time. I just came back from four weeks in China and Australia ^^

ngzhian:
fixed :)

koalaman:
Nice, thanks!

PR #1423, opened 2018-12-21
Fix applying multiple fixes per line

### LINKED ISSUE #1421: Autofix does not apply multiple fixes per line
#### For bugs
- Rule Id (if any, e.g. SC1000): n/a
- My shellcheck version (`shellcheck --version` or "online"): ecd61bfc6
- [ ] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2086)

```sh
#!/bin/sh
echo $1 and $2
```

```
In foo line 2:
echo $1 and $2
     ^-- SC2086: Double quote to prevent globbing and word splitting.
            ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
echo "$1" and $2
```

```
Did you mean:
echo "$1" and "$2"
```

ngzhian:
Fixes #1421

### DIFF
--- a/src/ShellCheck/Fixer.hs
+++ b/src/ShellCheck/Fixer.hs
@@ -14,18 +14,9 @@ applyFix fix fileLines =
         -- prereq: list is already sorted by start position
         removeOverlap [] = []
         removeOverlap (x:xs) = checkoverlap x xs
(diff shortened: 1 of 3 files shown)


koalaman:
Nice!

PR #1424, opened 2018-12-22
Autofix tab

ngzhian:
WIP patch to let you see my approach to fix the tab issue in applying fix. I have more detailed message in commit d1e8c40 regarding an off-by-one error in calculating real column counts from virtual column counts.

### DIFF
--- a/src/ShellCheck/Formatter/TTY.hs
+++ b/src/ShellCheck/Formatter/TTY.hs
@@ -153,10 +153,13 @@ showFixedString color comments lineNum fileLines =
         -- fixes for that single line. We can fold the fixes (which removes
         -- overlaps), and apply it as a single fix with multiple replacements.
         applicableComments -> do
-            let mergedFix = (fold . catMaybes . (map pcFix)) applicableComments
+            let mergedFix = (realignFix . fold . catMaybes . (map pcFix)) applicableComments
             -- in the spirit of error prone
             putStrLn $ color "message" "Did you mean: "
             putStrLn $ unlines $ fixedString mergedFix fileLines
+        where
+            realignFix f = f { fixReplacements = map fix (fixReplacements f) }
+            fix r = realign r fileLines
 
 fixedString :: Fix -> [String] -> [String]
 fixedString fix fileLines =
--- a/src/ShellCheck/Fixer.hs
+++ b/src/ShellCheck/Fixer.hs
@@ -16,20 +16,38 @@ class Ranged a where
             yEnd = end y
             xStart = start x
             xEnd = end x
+    -- Set a new start and end position on a Ranged
+    setRange :: (Position, Position) -> a -> a
+
+instance Ranged PositionedComment where
+    start = pcStartPos
+    end = pcEndPos
+    setRange (s, e) pc = pc {
+        pcStartPos = s,
+        pcEndPos = e
+    }
 
 instance Ranged Replacement where
     start = repStartPos
     end   = repEndPos
+    setRange (s, e) r = r {
+        repStartPos = s,
+        repEndPos = e
+    }
 
 instance Ranged a => Ranged [a] where
     start [] = newPosition
     start xs = (minimum . map start) xs
     end []   = newPosition
     end xs   = (maximum . map end) xs
+    setRange (s, e) rs = map (setRange (s, e)) rs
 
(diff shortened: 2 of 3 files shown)


src/ShellCheck/Interface.hs
```diff
@@ -270,3 +274,36 @@ mockedSystemInterface files = SystemInterface {
             [] -> return $ Left "File not included in mock."
             [(_, contents)] -> return $ Right contents
 
+-- The Ranged class is used for types that has a start and end position.
```
koalaman: Btw I moved these typeclasses into `ShellCheck.Fixer` so that hopefully no one else will start depending on them

src/ShellCheck/Formatter/TTY.hs
```diff
@@ -147,17 +148,49 @@ showFixedString color comments lineNum fileLines =
...
-
-fixedString :: PositionedComment -> [String] -> [String]
-fixedString comment fileLines =
-    case (pcFix comment) of
-    Nothing -> [""]
-    Just fix -> case (fixReplacements fix) of
+            putStrLn $ unlines $ fixedString (makeNonVirtual2 mergedFix fileLines) fileLines
+
+-- Realign replacements from a tabstop of 8 to 1
+makeNonVirtual2 f ls =
```
koalaman: This approach would definitely work. 

I was thinking though.. what if we had a data structure that would represent a string, but also keep track of the original string indices.

For example, in linear form, it could be a `[(Int, String)]` where the int is the length of the piece, while the String is the contents. A trivial way to turn a string into such a form is e.g. `[(5, "hello)]`, or `[(1, "h"), (1, "e"), (1, "l"), (1, "l"), (1, "o")]`.

If you e.g. wanted to add a quote at index 0 and 5, you could find the index by counting, and then inserting to get `[(0, "\""), (5, "hello"), (0, "\"")]` .

The benefit of this is that when another fix wants to apply something at a given index, it can similarly count without having to adjust for previous string modifications. The other benefit is that by having `[(8, "\t")]`, tabs are automatically accounted for (at the start of the line anyways).

Realistically it would probably be in a tree form instead of a list for speed, with the same principle of letting you insert at original string indices.

Do you think such a thing could work?

ngzhian: Yea this looks interesting, how will it work with multiple lines, i.e. how will "hello\nworld" be represented?
how about deletions? What will deleting the first char from `(5, "hello")` look like?

ngzhian: I think for multiple lines it would be simpler to treat it as a single line containing new lines, and do the replacement logic for a single line. (a made-up problematic example would be if the fix is spans 3 rows but the range to be replaced spans 2 rows:
```
[ [(1, "h"), (2, "e")], [(1, "l"),(3, "lo")] ]
  -> [ [(1, "h"), (2, "e")], [(1, "l"),(1, "l")], [(1, "o")] ]
```
then for the last row `[(1, "o")]`, you probably need an indicator that it shouldn't be counted when counting line numbers for the next fix, (e.g. if a fix is for line 3, it's really the new line 4). So perhaps this structure will need a similar index for row, like it currently has for columns).

This is a good idea, having something more robust is good. But I don't have the knowledge or experience to know what the more robust thing looks like :) I'm leaning towards my crappy way of doing it, then filing a feature request with your notes for future implementation.

koalaman:
Are you still working on this or should I go ahead and merge?

ngzhian:
Let me clean this up a little bit :) I'll ping when it's ready to be reviewed, thanks!

ngzhian:
I have a Haskell question. I would like to change `makeNonVirtual` work on the `Ranged` typeclass that was created previously. `makeNonVirtual :: Ranged a => a -> [String] -> a`, but I have no idea how to *update* the start and end position within `Ranged `.

Right now `makeNonVirtual :: [PositionedComment] -> [String] -> [PositionedComment]` has access to the record selected (`pcStartPos`, `pcEndPos`), but I have no idea what to do in the case of a `Ranged a`.

Two workarounds is to have `makeNonVirtual` return 2 integers representing the new start and end column, or have `makeNonVirtual` take a setting function `Ranged a => a -> (Integer, Integer) -> a`.

ngzhian:
ready for review!

PR #1425, opened 2018-12-22
Add check for ambiguous nullary test

### LINKED ISSUE #1416: Encourage use of `[[ -n string ]]` rather than `[[ string ]]`

```sh
#!/bin/bash

if [[ $(true) ]]; then
  echo true
else
  echo false
fi 
```

No issues detected!

```
SCXXXX: If you did intend to check that the output of the command is
not an empty string, use the longer form of the `-n` conditional expression
to clarify the intent `[[ -n $butter ]]`. However, if you intend to check the 
exit code of the command then remove the conditional expression:

  if true; then
    echo true
  else
    echo false
  fi
```
The rationale for this feature request is that one arg version of -n, i.e.
`[[ $butter ]]` is not well known and some users confuse it with how you test
the exit code of a command. Encourage people to use `-n` which is easier
to search for in documentation, and makes the code easier to grok.

ngzhian:
Given an input like `if [[ $(a) ]]; then ...`, this is a implicit `-n` test,
so it works like `if [[ -n $(a) ]]; then ...`. Users might confuse this for
a check for the exit code of the command a, which should be tested with:

    if a; then
        ...

We warn the user to be more explicity and specifity the `-n`.

Fixes #1416

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -170,6 +170,7 @@ nodeChecks = [
     ,checkSubshelledTests
     ,checkInvertedStringTest
     ,checkRedirectionToCommand
+    ,checkNullaryExpansionTest
     ]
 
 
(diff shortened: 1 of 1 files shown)


src/ShellCheck/Analytics.hs
```diff
@@ -3096,5 +3097,19 @@ checkRedirectionToCommand _ t =
                 warn id 2238 "Redirecting to/from command name instead of file. Did you want pipes/xargs (or quote to ignore)?"
         _ -> return ()
 
+prop_checkNullaryExpansionTest1 = verify checkNullaryExpansionTest "[[ $(a) ]]"
+prop_checkNullaryExpansionTest2 = verify checkNullaryExpansionTest "[[ $a ]]"
+prop_checkNullaryExpansionTest3 = verifyNot checkNullaryExpansionTest "[[ $a=1 ]]"
+prop_checkNullaryExpansionTest4 = verifyNot checkNullaryExpansionTest "[[ -n $(a) ]]"
+checkNullaryExpansionTest _ t =
+    case t of
+        TC_Nullary _ _ (T_NormalWord _ (t2:[])) ->
```
koalaman: Might as well generalize this to any `T_NormalWord` to catch things like `[[ $a$b ]]`.

It could be gated on `not . isConstant`  instead of `isQuoteableExpansion` in that case.

src/ShellCheck/Analytics.hs
```diff
@@ -3096,5 +3097,19 @@ checkRedirectionToCommand _ t =
...
+            when (isQuoteableExpansion t2) $
+                warn (getId t2) 2243 (
+                "Use -n to check for null string, " ++
+                " or if you intend to check for the exit code of a command, " ++
```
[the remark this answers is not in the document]
ngzhian: I made 2243 for command substitution and 2244 for the rest.

src/ShellCheck/Analytics.hs
ngzhian: Followed your suggestion :)

koalaman:
Thanks! I added auto-fix support in 73822c3 (and also made it more resilient to quoting)

ngzhian:
Looks great, thank you!

ngzhian:
Hm, seems like we need to do variable flow analysis here? @koalaman

PR #1426, opened 2018-12-22
Update supported ulimit flags for dash

### LINKED ISSUE #1406: SC2169: "In dash, ulimit [...] is not supported" (but it is)
#### For bugs
- Rule Id: `SC2169`
- [X] I read the issue's wiki page, https://github.com/koalaman/shellcheck/wiki/SC2169
- [X] I tried on shellcheck.net and verified that this is still a problem

```sh

#!/bin/dash
ulimit -c unlimited
ulimit -n 100000
ulimit -Sp 3600

```

```
SC2169: In dash, ulimit -c is not supported.
SC2169: In dash, ulimit -n is not supported.
SC2169: In dash, ulimit -S is not supported.
```

No warning; in dash, `ulimit` commands **are** supported:

```
$ man dash
(search "ulimit")
ulimit [-H | -S] [-a | -tfdscmlpn [value]]
            Inquire about or set the hard or soft limits on processes or set new limits.  The choice between hard limit (which no process is allowed to violate, and which may not be raised once it has been
            lowered) and soft limit (which causes processes to be signaled but not necessarily killed, and which may be raised) is made with these flags:

            -H          set or inquire about hard limits

            -S          set or inquire about soft limits.  If neither -H nor -S is specified, the soft limit is displayed or both limits are set.  If both are specified, the last one wins.

            The limit to be interrogated or set, then, is chosen by specifying any one of these flags:

            -a          show all the current limits

            -t          show or set the limit on CPU time (in seconds)

            -f          show or set the limit on the largest file that can be created (in 512-byte blocks)

            -d          show or set the limit on the data segment size of a process (in kilobytes)

            -s          show or set the limit on the stack size of a process (in kilobytes)

            -c          show or set the limit on the largest core dump size that can be produced (in 512-byte blocks)

            -m          show or set the limit on the total physical memory that can be in use by a process (in kilobytes)

            -l          show or set the limit on how much memory a process can lock with mlock(2) (in kilobytes)

            -p          show or set the limit on the number of processes this user can have at one time

            -n          show or set the limit on the number files a process can have open at once

            -r          show or set the limit on the real-time scheduling priority of a process
```

Also, my script works without errors, so `dash` is definitely accepting these commands.

System info:
```
$ lsb_release -d
Description:	Linux Mint 18.3 Sylvia
```
(This is the same as Ubuntu 16.04 Xenial)

ngzhian:
Values are retrieved from https://linux.die.net/man/1/dash, search for
ulimit.

Fixes #1406

### DIFF
--- a/src/ShellCheck/Checks/ShellSupport.hs
+++ b/src/ShellCheck/Checks/ShellSupport.hs
@@ -136,6 +136,8 @@ prop_checkBashisms53= verifyNot checkBashisms "#!/bin/sh\nprintf -- -f\n"
 prop_checkBashisms54= verify checkBashisms "#!/bin/sh\nfoo+=bar"
 prop_checkBashisms55= verify checkBashisms "#!/bin/sh\necho ${@%foo}"
 prop_checkBashisms56= verifyNot checkBashisms "#!/bin/sh\necho ${##}"
+prop_checkBashisms57= verifyNot checkBashisms "#!/bin/dash\nulimit -c 0"
+prop_checkBashisms58= verify checkBashisms "#!/bin/sh\nulimit -c 0"
 checkBashisms = ForShell [Sh, Dash] $ \t -> do
     params <- ask
     kludge params t
@@ -283,7 +285,7 @@ checkBashisms = ForShell [Sh, Dash] $ \t -> do
             ("export", ["-p"]),
             ("printf", []),
             ("read", if isDash then ["r", "p"] else ["r"]),
-            ("ulimit", ["f"])
+            ("ulimit", if isDash then ["H", "S", "t", "f", "d", "s", "c", "m", "l", "p", "n"] else ["f"])
             ]
     bashism t@(T_SourceCommand id src _) =
         let name = fromMaybe "" $ getCommandName src


src/ShellCheck/Checks/ShellSupport.hs
```diff
@@ -283,7 +283,7 @@ checkBashisms = ForShell [Sh, Dash] $ \t -> do
             ("export", ["-p"]),
             ("printf", []),
             ("read", if isDash then ["r", "p"] else ["r"]),
-            ("ulimit", ["f"])
+            ("ulimit", ["H", "S", "t", "f", "d", "s", "c", "m", "l", "p", "n"])
```
koalaman: Nice! POSIX still only allows `f` though, so you can copy the pattern from the line above to make sure the warnings are still in place for `/bin/sh` (and maybe add one test case to verify that dash `ulimit -c 0` is fine, and one to make sure `sh` still gets a warning)

ngzhian: Updated, ptal :)

koalaman:
Nice, thanks!

PR #1430, opened 2018-12-28
Check jobs flags in dash/POSIX sh (fixes #1429)

### LINKED ISSUE #1429: In POSIX sh, jobs flags beside -l and -p are undefined

```sh
#!/bin/sh
jobs -s
jobs -r
```

No issues detected!

The only acceptable flags under POSIX sh for `jobs` are `-l` and `-p` ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/jobs.html)). The `-s` and `-r` flags are [valid on Bash](https://www.gnu.org/software/bash/manual/html_node/Job-Control-Builtins.html#Job-Control-Builtins) to check for stopped/suspended jobs and running jobs, but invalid on, for instance, dash:
```sh
$ jobs -s
dash: 1: jobs: Illegal option -s
$ jobs -r
dash: 2: jobs: Illegal option -r
```

I would expect shellcheck to warn under SC2039: "In POSIX sh, jobs flags beside -l and -p are undefined".

If you agree with the issue, I can submit a pull request adding the check.

contivero:
Most of the relevant info is already in #1429.


koalaman:
Thanks!

PR #1433, opened 2018-12-29
Fix 'export -p' being undefined under POSIX sh

### LINKED ISSUE #1432: False positive on POSIX sh with export -p
#### For bugs
- Rule Id: SC2039
- Shellcheck version 0.6.0 and online. 
- [x] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2039)

```sh
#!/bin/sh
export -p
```

```
Line 2:
export -p
       ^-- SC2039: In POSIX sh, export -p is undefined.
```

Nothing, `-p` is the only valid flag for export in POSIX sh ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#export)). This false positive was introduced when #1130 was fixed.

contivero:
Fixes #1432


koalaman:
Nice, thanks!

PR #1436, opened 2018-12-29
Fix off-by-one in overlap check for fixes

### LINKED ISSUE #1431: Autofix turns `cd $1` into `cd "$1 || exit"`
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2164, SC2086
- My shellcheck version (`shellcheck --version` or "online"): ef811995

```sh
#!/bin/sh
cd $1
```

```
In - line 2:
cd $1
^---^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
   ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
cd "$1 || exit"
```

```
In - line 2:
cd $1
^---^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
   ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
cd "$1" || exit

```

ngzhian:
Previously if we have 2 ranges like:
start: 6, end: 6
start: 6, end: 6
they would not be considered as overlap

Fixes #1431

### DIFF
--- a/src/ShellCheck/Fixer.hs
+++ b/src/ShellCheck/Fixer.hs
@@ -10,7 +10,7 @@ class Ranged a where
     end     :: a -> Position
     overlap :: a -> a -> Bool
     overlap x y =
-        (yStart >= xStart && yStart < xEnd) || (yStart < xStart && yEnd > xStart)
+        (yStart >= xStart && yStart <= xEnd) || (yStart < xStart && yEnd > xStart)
         where
             yStart = start y
             yEnd = end y


koalaman:
With 434b9047, most such "overlapping" zero-width spans will still be placed correctly thanks to tracking which AST node and side they were intended for, so this is no longer necessary. Thanks!

ngzhian:
This is awesome!! Great work!

Issue #1039, opened 2017-11-03
SC2221 "This pattern always overrides a later one" should show the patterns


bje-:
#### For bugs
- Rule Id SC2221 and SC2222
- shellcheck version: 0.4.6
- I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### For new checks and feature suggestions
- [X] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh

foo=123  # dummy
case $foo in
  foo) echo 1 ;;
  # many lines of cases
  foo) echo 2 ;;
esac
```

#### Here's what shellcheck currently says:
 
Line 5:
  foo) echo 1 ;;
  ^-- SC2221: This pattern always overrides a later one.
 
Line 7:
  foo) echo 2 ;;
  ^-- SC2222: This pattern never matches because of a previous pattern.

#### Here's what I wanted or expected to see:

It would be great if the duplicated pattern could be shown (and, perhaps the line number). eg:

Line 5:
  foo) echo 1 ;;
  ^-- SC2221: This pattern always overrides a later one (foo on line 7).
 
Line 7:
  foo*) echo 2 ;;
  ^-- SC2222: This pattern never matches because of a previous pattern (foo on line 5).

PR #1440, opened 2019-01-01
Add context to case pattern warnings

Gandalf-:
https://github.com/koalaman/shellcheck/issues/1039

This change adds line number references to the case patterns for SC2221 and SC2222

Builds using the default GHC and `stack build`. I'm not sure if there's a precedent for adding tests for these kinds of changes, let me know where to add some if needed.

## Testing
```shell
cat -n ~/code/shell/example.sh
     1  #!/bin/bash
     2
     3
     4  case $PATH in
     5    foo)
     6      echo yes
     7      ;;
     8    bar)
     9      echo
    10      ;;
    11    foo)
    12      echo no
    13      ;;
    14  esac
```

```
$ stack exec shellcheck ~/code/shell/example.sh

In /home/leaf/code/shell/example.sh line 5:
  foo)
  ^-^ SC2221: This pattern always overrides a later one on line 11.

In /home/leaf/code/shell/example.sh line 11:
  foo)
  ^-^ SC2222: This pattern never matches because of a previous pattern on line 5.

For more information:
  https://www.shellcheck.net/wiki/SC2221 -- This pattern always overrides a l...
  https://www.shellcheck.net/wiki/SC2222 -- This pattern never matches becaus...
```

```
leaf@aspen ~/l/shellcheck (issue_1039_case_pattern_context=)> stack test
...
Completed 2 action(s).
```


koalaman:
Looks good, thanks!

Issue #1404, opened 2018-12-01
SC2063 should not apply to grep --include '*.sh'?


hrvoj3e:
```
grep -r --include '*.sh' -L "something" .
                   ^ This is OK - GLOB is allowed
```

> http://linuxcommand.org/lc3_man_pages/grep1.html
> ```
> --include=GLOB
>      Search  only  files whose base name matches GLOB (using wildcard
>      matching as described under --exclude).
> ```

koalaman:
You're right, of course. This issue was kindly fixed by @Gandalf-  in 4a2b2c7, thanks!

PR #1444, opened 2019-01-07
Issue 1404 grep glob false positives

Gandalf-:
This addresses Issue https://github.com/koalaman/shellcheck/issues/1404

Some grep flags support globs; these are now all checked to prevent false positives. Unit test cases for each flag were added.

# Testing
```
I leaf@aspen ~/l/shellcheck (issue_1404_grep_globs=)> stack test
...

Completed 2 action(s).
```


koalaman:
Nice, thanks!

PR #1447, opened 2019-01-08
Fix #1340 (SC2093 about removing "exec" should trigger in loops)

### LINKED ISSUE #1340: SC2093 about removing "exec" should trigger in loops
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2093
- [x] I read the issue's wiki page, e.g. https://github.com/koalaman/shellcheck/wiki/SC2086

- [ ] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [ ] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

```sh
#!/bin/sh
for file in rc.d/*
do
     exec "$file"
done
```

No issues

SC2093: Remove "exec " if script should continue after this command.

ShellCheck does not emit this warning when the command is the last in its block, but it should still be triggering for loops.

jabberabbe:
Fixes issue #1340 
I implemented this adding a `Bool` argument, but if you prefer I can separate the two functions.
I added a property using the same code you gave in the issue.
`stack test` passes cleanly.


koalaman:
Looks good to me. Thanks!

PR #1449, opened 2019-01-10
Check readonly flags in dash/POSIX sh (fixes #1448)

### LINKED ISSUE #1448: In POSIX sh, readonly flags besides -p are undefined

```sh
#!/bin/sh
foo() { echo bar; }
readonly -f foo
```

No issues detected!

The only acceptable flag under POSIX sh for `readonly` is -p ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#readonly)). The same is true for dash, but bash also accepts [-aAf]. I expected to see the usual "In POSIX sh, x flags are undefined".

contivero:
Closes #1448


koalaman:
Nice, thanks!

PR #1456, opened 2019-01-13
Silence SC2103 when using 'set -e' (fixes #667)

### LINKED ISSUE #667: SC2103 not silenced by `set -e`
Using `set -e` silences [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164), on the grounds that with `set -e`, the `cd` command cannot fail. However [SC2103](https://github.com/koalaman/shellcheck/wiki/SC2103) is not silenced by `set -e`, despite the rationale being effectively the same.

contivero:
Hopefully that's enough. I followed how [SC2164 is being silenced](https://github.com/koalaman/shellcheck/blob/ec6f9e4d499c50685dc6caffcc3d4eb53320904a/src/ShellCheck/Analytics.hs#L2580).


koalaman:
This check should probably be fixed to not suggest adding a subshell around anything that modifies variables, but this is a good patch for now. Thanks!

PR #1458, opened 2019-01-14
Check cd flags under dash & POSIX sh (fixes #1457)

### LINKED ISSUE #1457: In POSIX sh, cd flags besides -L and -P are undefined

```sh
#!/bin/sh
set -e
cd -P -e dir
```

Nothing

"In POSIX sh, cd -e is undefined"

The only allowed flags under POSIX sh for the`cd` builtin are `-L` and `-P` ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cd.html)). Dash also follows this. Bash, however, allows: `[-L|[-P [-e]] [-@]`.

contivero:
Closes #1457


koalaman:
Thanks!

PR #1462, opened 2019-01-16
Check trap flags under dash & POSIX sh (fixes #1461)

### LINKED ISSUE #1461: In POSIX sh, trap flags are undefined

```sh
#!/bin/sh
trap -l
```

Nothing

"In POSIX sh, trap flags are undefined"

There are no flags for the `trap` builtin defined under POSIX sh ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#trap)), and the same is true for dash, but bash defines the flags [-lp].

contivero:
Closes #1461


koalaman:
Nice, thanks!

Issue #1369, opened 2018-10-21
Use file extension to detect shell


ingydotnet:
- Rule Id 2148
- My shellcheck version 0.5.0: 
- [x] I read the issue's wiki page, e.g. https://github.com/koalaman/shellcheck/wiki/SC2148
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

Why doesn't shellcheck consider the file extension to decide which shell is in play? I couldn't find any issues related to this. I write a lot of bash libraries that have a `.bash` file extension. I don't have a good reason for a shebang in those files.

jasonkarns:
👍 I agree. I typically follow the convention that executable scripts have a shebang, and sourced scripts have the appropriate .sh or .bash extension.

I would prefer to not add a shebang to scripts intended to be sourced, because it implies the scripts can be executed directly, which is not the case. I would love to see file extension used as a fallback for shell detection when a shebang isn't present.

ingydotnet:
Perhaps we can learn from vim here, which considers (at least):
* File extension
* Shebang line
* Modeline
  * eg `# vim: ft=sh:`

ie It might be valuable to also support a shellcheck directive like `# shellcheck shell=ksh` that has the final say.

matthewpersico:
Um, https://github.com/koalaman/shellcheck/wiki/Directive#shell

ingydotnet:
Oh cool. Thanks @matthewpersico.

That's not in `man shellcheck` for my 0.4.6 (and the wiki says it was added in 0.4.5). Only `disable` and `source` are listed as valid directives. Not sure if that's been fixed...

matthewpersico:
Yeah,@ingydotnet, gotta get up to at least 0.5.0; lots of fixes and improvements over 0.4.x.

ingydotnet:
BTW, I use 0.4.6 (by default) because that's what `apt install shellcheck` installs (by default) on Ubuntu 18.04. I'm not sure who is responsible for the man page. Perhaps Debian or Ubuntu. The copyright date in the man page is 2012-2015, so obviously not up to date for 2018.

ingydotnet:
@matthewpersico sorry didn't see your last comment. I do also have a local install (via cabal) of 0.5.0 that some of my build systems use. It seems that cabal does not have man pages.

ngzhian:
See #1317 , looks like Debian Sid has 5.0

ingydotnet:
[...]

PR #1464, opened 2019-01-18
Fix issue #1369 (Use file extension to determine shell)

jabberabbe:
Now ShellCheck can determine the shell type from the file extension if it is `.bash`,`.ksh` or `.dash`.
I have also changed the informative text for SC2148 to document the new feature. I'll modify the wiki page today.

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -484,7 +484,7 @@ checkShebang params (T_Annotation _ list t) =
 checkShebang params (T_Script id sb _) = execWriter $ do
     unless (shellTypeSpecified params) $ do
         when (sb == "") $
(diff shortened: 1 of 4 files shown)


koalaman:
I merged it and added some tests in a504ca6b5. Thanks!

I left SC2148 alone since it similarly doesn't mention the `-s` or directive options, and anyone who wants a non-standard, non-OS recognized way of specifying the shell can read all about it on the now-updated wiki page.

jabberabbe:
Sorry, I forgot something: shouldn't we edit `shellcheck.1.md`, in particular the part on the `-s` flag?

koalaman:
Good point, I've added this in 6dcf4b8e

Issue #1393, opened 2018-11-17
Suggestion: --quiet flag


l0b0:
#### For new checks and feature suggestions
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### The problem:

When trying to quickly determine whether there is any problem with any of a possibly large amount of files it would be very useful if `shellcheck` could

1. print nothing (or at least not to standard output) and
1. `exit 1` as soon as any issue is detected.

This would be equivalent to grep's `--quiet`, which also prints nothing and exits as soon as a match is found.

koalaman:
Thanks to @Gandalf-  for adding `-f quiet` for this!

PR #1465, opened 2019-01-19
Issue 1393 quiet flag

Gandalf-:
Issue https://github.com/koalaman/shellcheck/issues/1393

Provide '-q' and '--quiet' flags that suppress all normal output, but keep the
return status, similar to 'grep -q'. 

- new help text
```
$ stack exec -- shellcheck --help
Usage: shellcheck [OPTIONS...] FILES...
  -a                --check-sourced          Include warnings from sourced files
  -C[WHEN]          --color[=WHEN]           Use color (auto, always, never)
  -e CODE1,CODE2..  --exclude=CODE1,CODE2..  Exclude types of warnings
  -f FORMAT         --format=FORMAT          Output format (checkstyle, gcc, json, tty)
  -s SHELLNAME      --shell=SHELLNAME        Specify dialect (sh, bash, dash, ksh)
  -S SEVERITY       --severity=SEVERITY      Minimum severity of errors to consider (error, warning, info, style)
  -q                --quiet                  Suppress all normal output
  -V                --version                Print version information
  -W NUM            --wiki-link-count=NUM    The number of wiki links to show, when applicable
  -x                --external-sources       Allow 'source' outside of FILES
                    --help                   Show this usage summary and exit
```

# Testing
```
$ stack test
...
=== prop_readScript5 from src/ShellCheck/Parser.hs:2975 ===
+++ OK, passed 100 tests.

Completed 2 action(s).
```
- example normal behavior, all output show, exit code set
```
$ cat /tmp/example.sh
#!/bin/bash

while [[ $1 ]]; do
  echo "$1"
  shift
done

$ stack exec -- shellcheck /tmp/example.sh

In /tmp/example.sh line 3:
while [[ $1 ]]; do
         ^-- SC2244: Prefer explicit -n to check non-empty string (or use =/-ne to check boolean/integer).

Did you mean:
while [[ -n $1 ]]; do

For more information:
  https://www.shellcheck.net/wiki/SC2244 -- Prefer explicit -n to check non-e...

$ echo $?
1
```
- quiet mode behavior, no output, exit code set
```
$ stack exec -- shellcheck --quiet /tmp/example.sh
$ echo $?
1
```


Gandalf-:
I initially missed updating `shellcheck.1.md`, that's included now

koalaman:
I can see this being useful.

I have a suggestion for a different approach though: how about adding a `Quiet` output formatter?

It could just have no-ops for headers/footers, and its `onResult` function could just exit immediately if there had been anything to report. That way, it won't process more files than necessary.

Gandalf-:
I'll give that a shot, it sounds like a cleaner approach. Thanks for the feedback!

Gandalf-:
# With issues
- usual behavior
```
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> cat -n /tmp/example.sh
     1  #!/bin/bash
     2
     3  while [[ $1 ]]; do
     4    echo "$1"
     5    shift
     6  done
```
```
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> stack exec -- shellcheck /tmp/example.sh

In /tmp/example.sh line 3:
while [[ $1 ]]; do
         ^-- SC2244: Prefer explicit -n to check non-empty string (or use =/-ne to check boolean/integer).

Did you mean:
while [[ -n $1 ]]; do

For more information:
  https://www.shellcheck.net/wiki/SC2244 -- Prefer explicit -n to check non-e...
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> echo $status
1
```
- quiet
```
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> stack exec -- shellcheck -f quiet /tmp/example.sh
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> echo $status
1
```
# Without issues
```
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> cat -n /tmp/example.sh
     1  #!/bin/bash
     2
     3  var=1
     4  echo $var
```
```
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> stack exec -- shellcheck /tmp/example.sh
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> echo $status
0
```
- quiet 
```
I leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> stack exec -- shellcheck --format quiet /tmp/example.sh
N leaf@willow ~/w/shellcheck (issue_1393_quiet_flag=)> echo $status
0
```

Gandalf-:
I tested the changes on files with just suggestions (green), just warnings (yellow), and just errors (red), and a mix - it returned 1 with each. 

I feel like there might be something missing with `onResult` as a no-op, but it appears to behave correctly

src/ShellCheck/Formatter/Quiet.hs
```diff
@@ -0,0 +1,36 @@
+{-
+    Copyright 2012-2015 Vidar Holen
```
koalaman: This is yours now ^^

src/ShellCheck/Formatter/Quiet.hs
```diff
@@ -0,0 +1,36 @@
...
+import System.Exit
+
+format :: FormatterOptions -> IO Formatter
+format options = do
+    topErrorRef <- newIORef []
+    return Formatter {
+        header = return (),
+        footer = return (),
+        onFailure = \ _ _ -> exitFailure,
+        onResult  = \ _ _ -> return ()
```
koalaman: `onFailure` is called when the file can't be checked, e.g. due to no permissions, so that one's good.

`onResult` is called when the file can be checked, including when it has suggestions, so you can speed this up further with e.g. `onResult = \result _ -> unless (null $ crComments result) exitFailure`.

It'll still give the correct result right now, but it'll scan all the files instead of exiting on the first one with suggestions.

koalaman:
Looks good, thanks!

Issue #417, opened 2015-07-23
bats support


aw-was-here:
It would be nice if shellcheck supported/understood bats' @test nomenclature so that we could run shellcheck over bats input files.

nicolaiskogheim:
+1

koalaman:
What would be required to do this?

* Parse @test as brace groups
* Ignore the variables `status`, `lines` and `output`
* Assume `bats` in the shebang means `bash`

Am I missing anything?

aw-altiscale:
I think that's it.  If nothing else, it's a good start and the blanks can get filled in as we hit them.
 Awesome thanks!

bertvv:
+1!

Skilgarriff:
Any update on this? Otherwise a point in the right direction as to where we could help make this happen would be awesome!

dougm:
Using bats-preprocess works well and preserves the original .bats file line numbers, column numbers, etc:

```
/usr/local/libexec/bats-preprocess < foo.bats | shellcheck -
```

dougm:
Using this approach via Emacs + flycheck works well too: https://github.com/dougm/bats-mode/commit/001187c7544085c62c0141b380b43d8d77a497fb

Skilgarriff:
@dougm I was looking for a solution that would eventually then be able to port to Atom - Plus it would be nice to be able to be checking Bats formatting with shellcheck instead of first processing it to bash.

dougm:
@Skilgarriff agreed, it would be ideal for shellcheck to have native bats support.

koalaman:
I took a crack at this, but it seems to have more edge cases than I imagined.

I implemented the outlined additions, but bats apparently has a whole little DSL for including files and such that I didn't attempt to duplicate. It seems to work fine with bats' own tests. 

If anyone wants to try it out, check the `bats` branch: https://github.com/koalaman/shellcheck/tree/bats

edouard-lopez:
related:
* [sstephenson/bats: Why BATS syntax raises so many shellsheck errors?](https://github.com/sstephenson/bats/issues/188)
* [AtomLinter/linter-shellcheck: Why BATS syntax raises so many shellsheck errors?](https://github.com/AtomLinter/linter-shellcheck/issues/84)

zmarouf:
I just tested the ```bats``` branch with a few of my files. It works great :) Thanks!
(Note: I don't use the ```load```  feature extensively)

edouard-lopez:
[...]

PR #1469, opened 2019-01-20
Add bats support

damienrg:
This is motivated by the fact that the popularity of bats is increasing
since the creation of bats-core/bats-core.

The code is a cherry-pick of [koalaman/shellcheck/bats](/koalaman/shellcheck/tree/bats/) branch.

Fix koalaman/shellcheck#417.


koalaman:
Wow, 1 in every 89 files from my pile of GitHub scripts is now bats. This is certainly overdue. Thanks!

I additionally fixed resulting warnings about `local` in bats tests and added `.bats` to the freshly merged list of recognized extensions in e1fe9be7

PR #1472, opened 2019-01-21
Check type flags under dash and POSIX sh (fixes #1471)

### LINKED ISSUE #1471: In POSIX sh, type flags are undefined

```sh
#!/bin/sh
type -a ls
```

Nothing

"In POSIX sh, type flags are undefined"

There are no flags for the `type` builtin defined under POSIX sh ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/type.html)), and the same is true for dash. Bash, however, allows [-aftpP].

contivero:
Fixes #1471


koalaman:
Thanks!

Issue #1330, opened 2018-09-05
False positive on echo


GoldsteinE:
#### For bugs
- Rule Id: SC2039
- version 0.3.7
- [x] I read the issue's wiki page, e.g. https://github.com/koalaman/shellcheck/wiki/SC2086
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/sh
echo "-e hello"
```

#### Here's what shellcheck currently says:
```In - line 2:
echo "-e hello"
     ^-- SC2039: In POSIX sh, echo flags are not supported.
```

#### Here's what I wanted or expected to see:
No error. "-e hello" is not a flag, it's a string with space.

bittorf:
you can avoid that (and this is common) by using:

echo -- "-e foo"

GoldsteinE:
No, I totally can't.
```
sh-4.3$ echo -- "-e foo"
-- -e foo```
How correctly says shellcheck,
> In POSIX sh, echo flags are not supported
So `-- -e foo` will be outputted literally.

bittorf:
sorry about my mistake, you are correct and this needs a fix.

koalaman:
This is an interesting case.

In POSIX, the command works perfectly fine because no options (not even `--`) shall be recognized. This means that it shouldn't really be silenced by `--`.  The advice should probably be to use `printf` whenever arguments start with a `-`.

In Bash, it works because `"-e hello"` is still arguably a set of flags, but since it contains currently unsupported options (`-hlo `) it's being printed verbatim:

    $ echo "-Eeenenenenenenne" # BATMAN
    (no output)

    $ echo "-Eeenenenenenennex"
    -Eeenenenenenennex

Though of course there is no warning there.

koalaman:
Thanks to @Gandalf- , this now only warns when the argument consists entirely of known `echo` flags (specifically `-eEsn`)

PR #1473, opened 2019-01-22
Issue 1330 unsupported echo flags

Gandalf-:
Issue https://github.com/koalaman/shellcheck/issues/1330

Addresses false positives when quoted arguments to echo begin with what
looks like a flag. Now, warn only when the first argument is a
recognized echo flag and when flags are unsupported.

Second attempt, recognizes multiple flags together using regex. New test cases to verify.

## Testing
```
I leaf@willow ~/w/shellcheck (issue_1330_unsupported_echo_flags=)> stack test
...
```

### DIFF
--- a/src/ShellCheck/Checks/ShellSupport.hs
+++ b/src/ShellCheck/Checks/ShellSupport.hs
@@ -152,6 +152,10 @@ prop_checkBashisms69 = verifyNot checkBashisms "#!/bin/sh\numask -S"
 prop_checkBashisms70 = verify checkBashisms "#!/bin/sh\ntrap -l"
 prop_checkBashisms71 = verify checkBashisms "#!/bin/sh\ntype -a ls"
 prop_checkBashisms72 = verifyNot checkBashisms "#!/bin/sh\ntype ls"
+prop_checkBashisms73 = verifyNot checkBashisms "#!/bin/sh\necho \"-n foo\""
+prop_checkBashisms74 = verifyNot checkBashisms "#!/bin/sh\necho \"-ne foo\""
+prop_checkBashisms75 = verifyNot checkBashisms "#!/bin/sh\necho -Q foo"
+prop_checkBashisms76 = verify checkBashisms "#!/bin/sh\necho -ne foo"
 checkBashisms = ForShell [Sh, Dash] $ \t -> do
     params <- ask
     kludge params t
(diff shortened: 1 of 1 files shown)


src/ShellCheck/Checks/ShellSupport.hs
```diff
@@ -238,15 +242,19 @@ checkBashisms = ForShell [Sh, Dash] $ \t -> do
...
                 then
                     when (argString /= "-n") $
                         warnMsg (getId arg) "echo flags besides -n"
                 else
                     warnMsg (getId arg) "echo flags are"
-      where argString = concat $ oversimplify arg
+      where
+          argString = concat $ oversimplify arg
+          flagRegex = mkRegex "^-[eEsn]+$"
+          hasEchoFlags = Just [] == matchRegex flagRegex argString
```
koalaman: You can simplify this with the function `matches :: String -> Regex -> Bool`, commonly seen with infix notation:  ``argString `matches` matchRegex``

src/ShellCheck/Checks/ShellSupport.hs
```diff
@@ -238,15 +242,19 @@ checkBashisms = ForShell [Sh, Dash] $ \t -> do
         warnMsg id "`<file` to read files is"
 
     bashism t@(T_SimpleCommand _ _ (cmd:arg:_))
-        | t `isCommand` "echo" && "-" `isPrefixOf` argString =
+        | t `isCommand` "echo" && hasEchoFlags =
             unless ("--" `isPrefixOf` argString) $ -- echo "-----"
```
koalaman: I guess this part is redundant now

src/ShellCheck/Checks/ShellSupport.hs
```diff
@@ -238,15 +242,19 @@ checkBashisms = ForShell [Sh, Dash] $ \t -> do
...
                 if isDash
                 then
                     when (argString /= "-n") $
                         warnMsg (getId arg) "echo flags besides -n"
                 else
                     warnMsg (getId arg) "echo flags are"
-      where argString = concat $ oversimplify arg
+      where
+          argString = concat $ oversimplify arg
```
koalaman: You could optionally replace `concat $ oversimplify arg` with the less heavy handed version `getLiteralString arg`. This returns a `Maybe String`, with Nothing if the value is not purely literal, such as if it contains a variable (`oversimplify` instead arbitrarily inserts `"${VAR}"` in place of the variable)

Alternatively, just leave it.

Gandalf-:
Thanks for the feedback! I'll update the commit this evening.

Gandalf-:
# Testing
```
I leaf@willow ~/w/shellcheck (issue_1330_unsupported_echo_flags=)> stack test
...

```

koalaman:
Nice! This got rid of a bunch of false positives.

PR #1474, opened 2019-01-22
Check unset flags under dash and POSIX sh

contivero:
The only acceptable flags for the `unset` builtin under POSIX sh and dash
are [-fv] ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#unset)). Bash though, accepts [-n] too. This commits makes shellcheck
warn about this.


src/ShellCheck/Checks/ShellSupport.hs
```diff
@@ -306,7 +308,8 @@ checkBashisms = ForShell [Sh, Dash] $ \t -> do
             ("trap", Just []),
             ("type", Just []),
             ("ulimit", if isDash then Nothing else Just ["f"]),
-            ("umask", Just ["S"])
+            ("umask", Just ["S"]),
+            ("unset", ["f", "v"])
```
koalaman: I updated this part to allow not checking flags (because `ulimit`'s list for Dash was getting really long and it was missing some options, so it seemed better not try to check in that case).

All that's needed is a `Just` in front of `["f", "v"]`

contivero: I had somehow forgotten to commit the change, so it compiled and passed tests locally, but I pushed an outdated commit. Sorry for that.

koalaman:
Thanks!

PR #1480, opened 2019-01-28
Check set flags under dash & POSIX sh (fixes #990)

### LINKED ISSUE #990: Failure to detect illegal option set -o pipefail in POSIX sh (dash)
#### For bugs

```sh
[josh@kek ~ ]$ cat test.sh 
#!/usr/bin/env sh
set -eo pipefail
false | true
printf 'done\n'
[josh@kek ~ ]$ sh test.sh 
test.sh: 3: set: Illegal option -o pipefail
[josh@kek ~ ]$ shellcheck -s sh test.sh 
[josh@kek ~ ]$ echo $?
0
[josh@kek ~ ]$ shellcheck -V
ShellCheck - shell script analysis tool
version: 0.4.6
license: GNU General Public License, version 3
website: http://www.shellcheck.net
[josh@kek ~ ]$ which sh
/bin/sh
[josh@kek ~ ]$ ll /bin/sh
lrwxrwxrwx 1 root root 4 Jun 28 12:26 /bin/sh -> dash
```

Nothing, exit code 0.

Something like `Option pipefail not supported in (da)sh`.

contivero:
The `set` builtin accepts certain flags, and some longer synonyms with `-o` under POSIX sh. Dash diverges a bit from this (adds a couple of flags, and doesn't support `-h`). Taking this into account, this makes  ShellCheck warn if any of the used flags and options are undefined, based on whether the shebang is that of POSIX sh or dash.

This fixes #990, while adding general flag-support checking for `set` in the process.

### DIFF
--- a/src/ShellCheck/Checks/ShellSupport.hs
+++ b/src/ShellCheck/Checks/ShellSupport.hs
@@ -33,6 +33,7 @@ import Data.Char
 import Data.List
 import Data.Maybe
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 import Test.QuickCheck.All (forAllProperties)
 import Test.QuickCheck.Test (quickCheckWithResult, stdArgs, maxSuccess)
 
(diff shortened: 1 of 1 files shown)


koalaman:
Nice! I see this catches a number of `pipefail` violations.

There are some false positives though, such as when using `sh` style array building:

    set -- wget -o foo "$url"
    echo "Running $*"
    "$@"

It should stop looking for `set` options when it sees `--` or any word that doesn't start with a dash.

It also triggers on variables:

    opts=$-
    set -"$opts"

This is due to `oversimplify`, which (true to its name) pretends like the user just wrote `'-${VAR}'`. You can instead use `getLiteralString myword` which would return `Just "-eux"` for a literal string (regardless of its quoting), and Nothing if the string contains variables or other unknowns where it's probably better to trust the user. 

Since Dash already has a pile of options and are free to add new ones whenever they like, it might be better to limit this check to `sh`. People are going to check their scripts with two year old ShellCheck versions, so it's not just a matter of keeping it in sync.

contivero:
> It should stop looking for `set` options when it sees `--` or any word that doesn't start with a dash.

I agree that it should stop when finding `--`, but stopping on any word that doesn't start with a dash wouldn't catch for instance:
```sh
set -o errexit -o pipefail
``` 
because it would stop on `errexit`, or am I missing something?

>  You can instead use `getLiteralString myword`

I'll try with that and see if I can remove false positives, thanks!

> Since Dash already has a pile of options and are free to add new ones whenever they like, it might be better to limit this check to `sh`.

I thought you might bring this up, considering you removed the checks from `ulimit` too. Fair enough, I'll leave it for `sh` only.

PR #1488, opened 2019-02-10
Check wait flags in dash & POSIX sh

contivero:
Flags for the `wait` builtin are undefined under both POSIX sh ([see spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/wait.html)) and dash. Bash though, accepts `[-fn]`.


koalaman:
Thanks! Sorry it took a while.

Issue #896, opened 2017-05-03
printf -v 'array[0]' does not count as array assignment


koalaman:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2154
- My shellcheck version (`shellcheck --version` or "online"): online
- [X] I tried on shellcheck.net and verified that this is still a problem on the latest commit
- [ ] It's not reproducible on shellcheck.net, but I think that's because it's an OS, configuration or encoding issue

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/bash
var=howdy;
printf -v 'array[0]' %s "$var";
printf %s\\n "${array[@]}";
```

#### Here's what shellcheck currently says:

```
Line 4:
printf %s\\n "${array[@]}";
              ^-- SC2154: array is referenced but not assigned.
```
#### Here's what I wanted or expected to see:

Nothing

e-kwsm:
closed by #1499

Issue #433, opened 2015-08-11
printf -v and arrays


izabera:
printf doesn't set array variables correctly
```
printf -v "var[2]" x; echo "${var[2]}"
                            ^-- SC2154: var is referenced but not assigned.
```

kurahaupo:
Or rather, ```printf``` _does_ set array variables correctly, but shellcheck fails to note that it has done so.

e-kwsm:
closed by #1499

PR #1499, opened 2019-02-24
Fix issues #896 and #433: printf -v and arrays

jabberabbe:
Now ShellCheck recognises array variables assigned by `printf -v`.

# Testing
`stack build && stack test`
```
=== prop_checkUnassignedReferences37 from src/ShellCheck/Analytics.hs:1957 ===
+++ OK, passed 1 tests.
...
Completed 2 action(s).
```


koalaman:
I like that it doesn't just strip the index but also infers that it's an array. Very nice, thanks!

Issue #1523, opened 2019-03-17
encourage using 'basename' always with --


bittorf:
proposal:

```
#!/bin/sh
file='-/a/b/c.de'
basename "$file"
```

this throws an error:
```
# basename "$file"
basename: invalid option -- '/'
Try 'basename --help' for more information.
```

so better remind the use to always call with
```
basename -- "$file"
```

tripleee:
As already commented on the PR, this is by no means specific to `basename`, and so the check should probably be extended to any command which accepts a file name parameter. `basename` is a common victim to this because people do

```bash
for filename in *; do
    basename "$filename"
done
```
but the proper fix here is to say
```
for filename in ./*; do
```

PR #1532, opened 2019-03-24
Issue 1523 encourage basename '--' usage

Gandalf-:
Addresses issue https://github.com/koalaman/shellcheck/issues/1523

Checks if '--' should be the first argument to basename, to protect
against filenames being interpreted as options.

# Testing
```
I leaf@sitka ~/w/s/src (issue_1523_basename_dashes=)> stack build

I leaf@sitka ~/w/s/src (issue_1523_basename_dashes=)> echo $status
0
```
```
N leaf@sitka ~/w/s/src (issue_1523_basename_dashes=)> stack test
...
=== prop_checkBasename1 from src/ShellCheck/Checks/Commands.hs:500 ===
+++ OK, passed 1 tests.

=== prop_checkBasename2 from src/ShellCheck/Checks/Commands.hs:501 ===
+++ OK, passed 1 tests.

=== prop_checkBasename3 from src/ShellCheck/Checks/Commands.hs:502 ===
+++ OK, passed 1 tests.

=== prop_checkBasename4 from src/ShellCheck/Checks/Commands.hs:503 ===
+++ OK, passed 1 tests.

=== prop_checkBasename5 from src/ShellCheck/Checks/Commands.hs:504 ===
+++ OK, passed 1 tests.

=== prop_checkBasename6 from src/ShellCheck/Checks/Commands.hs:505 ===
+++ OK, passed 1 tests.

=== prop_checkBasename7 from src/ShellCheck/Checks/Commands.hs:506 ===
+++ OK, passed 1 tests.
...
I leaf@sitka ~/w/s/src (issue_1523_basename_dashes=)> echo $status
0

```

### DIFF
--- a/src/ShellCheck/Checks/Commands.hs
+++ b/src/ShellCheck/Checks/Commands.hs
@@ -69,6 +69,7 @@ commandChecks = [
     ,checkMkdirDashPM
     ,checkNonportableSignals
     ,checkInteractiveSu
+    ,checkBasename
     ,checkSshCommandString
     ,checkPrintfVar
     ,checkUuoeCmd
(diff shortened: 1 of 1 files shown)


koalaman:
This is a general problem that applies to most commands, not just `basename`. I think a check specific to `basename` will be too narrow. 

I'm not quite sure what the best course of action is though: requiring `--` for all commands may be excessively noisy, and doing it only for arguments starting with a dash is more likely to trigger for legitimate flags in variables.

Gandalf-:
Could we come up with a list of common commands where it'd have the most impact, and apply it to those? Not all commands even recognize `--` as an option

Perhaps awk, grep, cat, tar, bash and sh?

leagris:
Maybe with all commands that deals with a file-path argument, it should warn when the argument comes from a variable, but accept literal file-paths in single quotes without a warning.

A possibility is to both use a hard-coded list of the POSIX and GNU CoreUtils commands accepting a FILE argument, and otherwise brute-forcing the actual command content if it exist in the `$PATH` and is executable by the current user, like I did in this script:

```Bash
#!/usr/bin/env bash
command_has_FILE_argument() {
  (($#)) || return 1
  grep --silent --fixed-strings '[FILE' < <(strings "${1}")
}

sort --unique < <(
  IFS=$':'
  for path in $PATH; do
    while IFS= read -r -d '' command_path; do
      if command_has_FILE_argument "${command_path}"; then
        basename -- "${command_path}"
      fi
    done < <(find "${path}" -executable -type f -print0)
  done
)
```

Gandalf-:
This PR doesn't seem to fully address the issue, I'm going to close it for now. 

Could we come up with a list like @leagris suggested? Further discussion is probably best in: https://github.com/koalaman/shellcheck/issues/1523

Issue #510, opened 2015-10-16
Allow command line to define root for external-sources check


knirch:
It might|would be nice if something like

knirch@traktor:~/source/linux-dist$ shellcheck -x --root-path=$PWD -f gcc $(git grep -l '^#!/bin/bash')

would let shellcheck translate "source /usr/lib/shell/db-functions" to "source /home/knirch/linux-dist/usr/lib/shell/db-function"

I do know about the source=file directive, but it's going to add lots of noise; 
$ shellcheck -f gcc $(git grep -l '^#!/bin/bash' ) | grep -c SC1091
628

I could perhaps move some parts to dotfiles, but that would sadly remove the benefit of source checking.

koalaman:
With 301705edea you can now do this with `-P "$PWD"`, which changes the search path both for what's considered the root in absolute paths, and for the base path of relative paths.

This should probably also be part of the new `.shellcheckrc` functionality, but isn't yet.

(Happy 3.5 year anniversary for this issue!)

PR #1553, opened 2019-04-22
Add option to look for sources in alternate root paths

epontan:
Add a new optional flag `-r|--root ROOTPATHS`, where `ROOTPATHS` is a colon separated list of paths, that will look for external sources in alternate roots.

This is particular useful when the run-time environment does not fully match the development environment. The `#shellcheck source=file` directive is useful, but has its limitations in certain scenarios. Also, in many cases the directive could be removed from scripts when the root flag is used.

Script example.bash:
```bash
#!/bin/bash
source /etc/foo/config
```

Example usage where `etc/foo/config` exists in `skel/foo`:
```bash
shellcheck -x -r skel/foo:skel/core example.bash
```

### DIFF
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -69,6 +69,7 @@ instance Monoid Status where
 data Options = Options {
     checkSpec        :: CheckSpec,
     externalSources  :: Bool,
+    rootPaths        :: [FilePath],
     formatterOptions :: FormatterOptions,
     minSeverity      :: Severity
 }
@@ -76,6 +77,7 @@ data Options = Options {
 defaultOptions = Options {
     checkSpec = emptyCheckSpec,
     externalSources = False,
+    rootPaths = [],
     formatterOptions = newFormatterOptions {
         foColorOption = ColorAuto
     },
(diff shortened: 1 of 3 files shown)


austin987:
The help and your comment say semi colon, but the code is using a colon for delimiter.

epontan:
> The help and your comment say semi colon, but the code is using a colon for delimiter.

Thanks for noticing that! I don't know why I wrote semi-colon when I really was meaning colon :)
Anyway, it is fixed now.

knirch:
This would resolve #510

koalaman:
Very nice, thanks! 

I expanded this a bit into "source paths" (`-P`):

* For absolute paths, it behaves the same
* It also looks up relative paths based on these dirs  (so `-P "$PATH"` will find `. foo` in any path entry)
* It accepts a special name `SCRIPTDIR` to allow you to search relative to the current script's directory
* It uses platform canonical search path parsing, so that `c:\foo;d:\bar` works as expected on Windows
* It allows you to specify multiple `-P` flags that accumulate,  e.g. with `export SHELLCHECK_OPTS='-P foo'; shellcheck -P bar ..`
* It checks whether the files are allowed by the sandboxing to avoid errors exposing which files exist (only really matters for shellcheck.net)

blueyed:
> * It accepts a special name `SCRIPTDIR` to allow you to search relative to the current script's directory

I suggest using something more explicit here - since "SCRIPTDIR" might be a valid path.
It might make sense to have "./" be handled as "local to script file" instead.

I.e. instead of `-P SCRIPTDIR/include` you would use `./include`, but `include` would still be relative to cwd.
Alternatively something like `::SCRIPTDIR::` could be used (i.e. something that contains characters not valid for filenames - would be `;;SCRIPTDIR;;` on Windows then.
I can see that this might be ugly, but wanted to mention this since it is easy to adjust before having it release now still.

PR #1560, opened 2019-04-29
Update brew install instructions for cabal-install

jamiekyle-eb:


koalaman:
The readme now suggests `haskell-for-mac` to avoid having to compile this pretty hefty package

PR #1561, opened 2019-04-29
New verbose check for braces around variable references

yetamrra:
This uses the new verbose warning level to add SC2250 as a new check to require braces around variable references.  Additional reasoning for this warning is in the individual commit message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,13 @@
 - Source paths: Use `-P dir1:dir2` or a `source-path=dir1` directive
                 to specify search paths for sourced files.
 - json1 format like --format=json but treats tabs as single characters
-- SC2249: Warn about `case` with missing default case (verbose)
-- SC2248: Warn about unquoted variables without special chars (verbose)
+- SC2250: Warn about variable references without braces (optional)
+- SC2249: Warn about `case` with missing default case (optional)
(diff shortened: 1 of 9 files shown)


koalaman:
I'm realizing that "verbose" is not suitably granular for especially divisive style choices (i.e. those I'm not personally a fan of :P)

It's probably time to add a `shellcheck enable=..` directive along with some discovery mechanism to allow more of these kinds of tests

yetamrra:
I was originally going to propose a new "off-by-default" warning category and a command-line option to enable specific warnings, but then you added verbose and it seemed like that would be mostly duplicate.  Do you want me to send the other approach as another PR for comparison?

Or since verbose is off by default, do you want me to just add the option?  Then passing -S verbose is essentially like -Wall from the compiler world, but most people can get away with "-S style --extra=1,2,3" to just enable a few of the verbose checks.

koalaman:
I've added such an option now! Would you mind rebasing on top of it?

Rather than iterating over get/set variables, you can just match against relevant AST nodes like in the other optional check `checkDefaultCase`. This makes it even simpler.

yetamrra:
Rebased.  Thanks for the tip about making it a node check; that cleans it up a lot!

koalaman:
Thanks!

frou:
Thank you for this check. I never knew I wanted it, but now that it's here I love it!

PR #1570, opened 2019-05-05
docs: README: Update pandoc command to match `Setup.hs`

virgilwashere:
# 📝 docs: Update README

## pandoc command to match `Setup.hs`

The sdist hook in [Setup.hs](Setup.hs) disables the `smart` extension
when creating man page.

```console
pandoc -s -f markdown-smart -t man shellcheck.1.md -o shellcheck.1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,8 +210,10 @@ or see the [storage bucket listing](https://shellcheck.storage.googleapis.com/in
 
 Distro packages already come with a `man` page. If you are building from source, it can be installed with:
 
-    pandoc -s -t man shellcheck.1.md -o shellcheck.1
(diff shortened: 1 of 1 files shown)


koalaman:
Well caught! According to GitHub, the "`markdown-smart` disables `--` as `en-dashes`." renders as a quotation, which doesn't seem right. I think you can just remove the explanation

virgilwashere:
From the [pandoc docs](https://pandoc.org/MANUAL.html#typography)

> ❗️ Note: If you are _writing_ Markdown, then the smart extension has the reverse effect: what would have been curly quotes comes out straight.

koalaman:
Thanks!

PR #1575, opened 2019-05-07
ensure docker invocation is ephemeral

efx:
Also adjust the tag to use the `:stable` tag mentioned in the prior
line.


README.md
```diff
@@ -195,7 +195,7 @@ From Docker Hub:
 
 ```sh
 docker pull koalaman/shellcheck:stable  # Or :v0.4.7 for that version, or :latest for daily builds
-docker run -v "$PWD:/mnt" koalaman/shellcheck myscript
+docker run --rm -v "$PWD:/mnt" koalaman/shellcheck:stable myscript
```
efx: I could not find a shorthand version of `--rm` in  the [`run`](https://docs.docker.com/engine/reference/commandline/run/) documentation.

koalaman:
Thanks!

Issue #1048, opened 2017-11-13
Normalize tabs for json formatter the same way we do for gcc


pwielgolaski:
#### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/usr/bin/env bash
	 [[ $n != 0 ]]
```

#### Here's what shellcheck currently says:
```
➜  sample shellcheck -f json issue.bash | jq
[
  {
    "file": "issue.bash",
    "line": 2,
    "endLine": 2,
    "column": 13,
    "endColumn": 13,
    "level": "warning",
    "code": 2154,
    "message": "n is referenced but not assigned."
  }
]
➜  sample shellcheck -f gcc issue.bash      
issue.bash:2:6: warning: n is referenced but not assigned. [SC2154]

```

#### Here's what I wanted or expected to see:

In gcc format we normalize tabs, could we do the same for json formatter, so we have the same column number in both?

koalaman:
The formatters follow the convention of the tools they mimic. `gcc` uses a tab size of 1, `checkstyle` uses a tab size of 8. Changing those would break compatibility.

There's no particular reason why the shellcheck JSON format couldn't have used the `gcc` style instead of `checkstyle` and `tty`, but it didn't, and I don't see any value in changing it now since that would break the integrations that rely on it (like Syntastic).

The tab size used by each formatter is documented: 
https://github.com/koalaman/shellcheck/wiki/Integration

pwielgolaski:
Make sense, I did not see it being documented.
It would make easier my integration with intellij, but I can handle it in code.
Btw. I implemented integration plugin.
https://github.com/pwielgolaski/shellcheck-plugin waiting for approval to get it available

blueyed:
Just for information: typically linters use the byte/character offset, and do not handle tab in a special way (therefore a tab in gcc has an offset of 1 I assume).

Regarding compatibility: Syntastic uses `-fgcc`, not the json formatter.

It is not trivial to adjust this in Vim, where you would have to look at the actual line and adjust the column offset based on the tabs before the reported column.

PR #1578, opened 2019-05-08
Allow variable tabstop widths

yetamrra:
Add a new -t option to specify the output width of tabstops, with a
default of 8 to match the current behavior.  Convert formatters that
currently use hardcoded 8-character tabstops (TTY and JSON) to respect
the new -t option.

This has two main use cases:
  1. Allow editors to pass -fjson -t1 so that they can consume the json
     output in a character-oriented way without breaking backwards
     compatibility.
  2. Allow users with narrow terminals to run
     shellcheck -tN | expand -tN
     to fit more levels of indentation on the screen.

Also addresses #1048.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
                    Enable with `-o` flags or `enable=name` directives
 - Source paths: Use `-P dir1:dir2` or a `source-path=dir1` directive
                 to specify search paths for sourced files.
+- json1 format like --format=json but treats tabs as single characters
 - SC2249: Warn about `case` with missing default case (verbose)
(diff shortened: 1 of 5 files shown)


koalaman:
I definitely like the first point.

Given that the option doesn't do much for most output modes and causes misaligned errors by default for tty output, what do you think of having this in the form of a new `json1` output format instead?

yetamrra:
I conceptually like -t because it's orthogonal to the choice of format, but I expect you're right that combining -t1 with json is likely to be the dominant use.  I'll rework this to a json1 format instead.

yetamrra:
Reworked to a json1 format instead.  Please take a look.

koalaman:
Nice, thanks!

PR #1592, opened 2019-05-27
SC2016: Don't trigger when using empty backticks

hugopeixoto:
When using '``' or '```', it should not suggest using double quotes.

My use case was printing slack code blocks:

```
echo '```'"$DIFF"'```'
```


koalaman:
Thanks!

Issue #1574, opened 2019-05-07
Is there any way where I can get a binary for macOS?


ignatov:
Hi @koalaman!

Thanks a lot for such a great tool!
I wonder if there any place where I can get a binary for macOS? I found builds for Linux and Windows, which is rather handy – https://github.com/koalaman/shellcheck#installing-the-shellcheck-binary
For macOS, I found only a https://homebrew.bintray.com/bottles/shellcheck-0.6.0_1.mojave.bottle.tar.gz

What do you think is it worth to build a binary for macOS on Travis for every single commit, as far as Travis supports macOS?

Gandalf-:
How would it be distributed though? brew? Or another MacOS package manager perhaps?

ignatov:
We'd like to download with IntelliJ-based products and as an external linter.

koalaman:
I don't see any reason not to. I don't think I'll be able to get to it myself any time soon, but contributions will certainly be considered

bt-thiago:
`shellcheck` is already available on homebrew:

```
$ brew info shellcheck
shellcheck: stable 0.6.0 (bottled), HEAD
Static analysis and lint tool, for (ba)sh scripts
https://www.shellcheck.net/
/usr/local/Cellar/shellcheck/0.6.0_1 (8 files, 7.2MB) *
  Poured from bottle on 2019-03-01 at 11:46:11
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/shellcheck.rb
==> Dependencies
Build: cabal-install ✘, ghc ✘, pandoc ✘
```

scriptingosx:
homebrew is great for individuals managing their own devices but is impossible to use if you have to centrally manage a fleet of dozens, hundreds or thousands of Macs. A pre-comiled macOS binary would already be a great help for MacAdmins. A downloadable pkg installer file which drops the spellcheck binary in `/usr/local/bin/` would be perfect.

scriptingosx:
I have a solution to build your own binary without brew here: https://scriptingosx.com/2019/07/build-a-shellcheck-installer-for-macos/

koalaman:
There are now macOS binaries built continuously. I have no idea how binary portability works for macOS, but they run on my macOS 10.14.6:

https://shellcheck.storage.googleapis.com/shellcheck-latest.darwin.x86_64.tar.xz

Please give them a try and let me know if there are any issues

PR #1594, opened 2019-05-29
Compiling a binary for macOS 

Luizm:
Hi there

I've looked for a binary for macOS, and saw this issue: https://github.com/koalaman/shellcheck/issues/1574

I'm not sure if it's the best option to compile binaries from macOS, but I did the test compiling the macOS running by Travis, but I'm not able to test the whole pipeline.

@koalaman Is it makes sense to you what I did and how could I proceed from here

### DIFF
--- a/.compile_binaries
+++ b/.compile_binaries
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+# This script compile shellcheck binaries
+set -ex
+
+# Remove all tests to reduce binary size
+./striptests
+
+mkdir -p deploy
+
+_cleanup(){
+  rm -rf dist shellcheck || true
+}
+
+if [ "$TRAVIS_OS_NAME" = 'linux' ]
+then
+  # Linux Docker image
+  name="$DOCKER_BASE"
+  DOCKER_BUILDS="$DOCKER_BUILDS $name"
+  docker build -t "$name:current" .
+  docker run "$name:current" --version
+  printf '%s\n' "#!/bin/sh" "echo 'hello world'" > myscript
+  docker run -v "$PWD:/mnt" "$name:current" myscript
+
+  # Copy static executable from docker image
+  id=$(docker create "$name:current")
+  docker cp "$id:/bin/shellcheck" "shellcheck"
+  docker rm "$id"
+  ls -l shellcheck
+  ./shellcheck myscript
+  for tag in $TAGS
+  do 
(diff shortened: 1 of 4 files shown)


.prepare_deploy
```diff
@@ -22,29 +22,42 @@ This binary was compiled on $(date -u).
...
+  done
+fi
+
+if [ "$TRAVIS_OS_NAME" = 'osx' ];
+then
+  for file in *.darwin-x86_64
+  do
+    base="${file%.*}"
+    cp "$file" "shellcheck"
+    tar -cJf "$base.darwin.x86_64.tar.xz" --transform="s:^:$base/:" README.txt LICENSE.txt shellcheck
```
koalaman: Will this `tar` command work on macOS?

.prepare_deploy
```diff
@@ -22,29 +22,42 @@ This binary was compiled on $(date -u).
...
-done
-
-for file in *.linux-armv6hf
-do
-  base="${file%.*}"
-  cp "$file" "shellcheck"
-  tar -cJf "$base.linux.armv6hf.tar.xz" --transform="s:^:$base/:" README.txt LICENSE.txt shellcheck
-  rm "shellcheck"
-done
+if [ "$TRAVIS_OS_NAME" = 'linux' ];
```
koalaman: You can just `shopt -s nullglob` instead of checking the OS to just let it skip any pattern where there are no matching outputs

.compile_binaries
```diff
@@ -0,0 +1,71 @@
...
+  _cleanup
+fi
+
+if [ "$TRAVIS_OS_NAME" = 'osx' ]; 
+then
+  # Darwin x86_64 static executable
+  sudo ln -s /usr/local/bin/gsha512sum /usr/local/bin/sha512sum
+  brew install cabal-install
+  cabal update
+  cabal install
```
koalaman: You can `cabal build shellcheck` and copy `dist/build/shellcheck/shellcheck` instead of installing it

.prepare_deploy
f0rmiga: It won't. To handle `--transform` in OS X, we will need to install `gnu-tar`. It then becomes available as `gtar`.

Homebrew is available in the OS X Travis:
https://docs.travis-ci.com/user/reference/osx/#homebrew

Luizm:
@koalaman Could you check it again?

koalaman:
Thanks!

Issue #1597, opened 2019-05-31
Feature request: shflags semantics


yetamrra:
#### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

[shflags](https://github.com/kward/shflags) is a popular library to process command-line flags (6000+ hits on github alone, plus hundreds more in Chromium OS).  It's not using any non-standard shell features, but scripts that use it produce a lot of shellcheck warnings because it creates variables at runtime and shellcheck doesn't know that's happening.  It also creates a few global variables that get warned about.

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/bash

. "/path/to/shflags"

DEFINE_string my_flag "def" "Help text"
DEFINE_int unused_flag 0 "Help text 2"
DEFINE_boolean bool_flag ${FLAGS_TRUE} "More help text"

FLAGS_HELP="Usage message goes here."

FLAGS "$@" || exit $?
eval set -- "${FLAGS_ARGV}"

echo "${FLAGS_my_flag}"
echo "${FLAGS_no_such_flag}"

[[ ${FLAGS_bool_flag} -eq ${FLAGS_FALSE} ]] || echo "Bool set"
```

#### Here's what shellcheck currently says:
```
In flags.sh line 7:
DEFINE_boolean bool_flag ${FLAGS_TRUE} "More help text"
                         ^-----------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
DEFINE_boolean bool_flag "${FLAGS_TRUE}" "More help text"


In flags.sh line 9:
FLAGS_HELP="Usage message goes here."
^--------^ SC2034: FLAGS_HELP appears unused. Verify use (or export if used externally).


In flags.sh line 14:
echo "${FLAGS_my_flag}"
      ^--------------^ SC2154: FLAGS_my_flag is referenced but not assigned.


In flags.sh line 15:
echo "${FLAGS_no_such_flag}"
      ^-------------------^ SC2154: FLAGS_no_such_flag is referenced but not assigned.


In flags.sh line 17:
[[ ${FLAGS_bool_flag} -eq ${FLAGS_FALSE} ]] || echo "Bool set"
   ^----------------^ SC2154: FLAGS_bool_flag is referenced but not assigned.
```

This has three shortcomings:

1. It misses that unused_flag isn't used anywhere.
[...]

PR #1604, opened 2019-06-07
Add support for basic shflags semantics

yetamrra:
The shflags command-line flags library creates variables at runtime
with a few well-defined functions.  This causes shellcheck to spit out
lots of warnings about unassigned variables, as well as miss warnings
about unused flag variables.

We can address this with two parts:

1. Pretend that the shflags global variables are predefined like other
   shell variables so that shellcheck doesn't expect users to set them.
2. Treat DEFINE_string, DEFINE_int, etc. as new commands that create
   variables, similar to the existing read, local, mapfile, etc.

Part 1 can be addresssed without this by following sourced files, but
that doesn't help if people are otherwise not following external
sources.

The new behavior is turned on by a new --uses-shflags command-line
parameter.

Addresses #1597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Source paths: Use `-P dir1:dir2` or a `source-path=dir1` directive
                 to specify search paths for sourced files.
 - json1 format like --format=json but treats tabs as single characters
+- Recognize FLAGS variables created by the shflags library.
 - SC2154: Also warn about unassigned uppercase variables (optional)
 - SC2252: Warn about `[ $a != x ] || [ $a != y ]`, similar to SC2055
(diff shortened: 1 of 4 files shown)


koalaman:
Nice! I think it would be fine to just always have `DEFINE_string` etc define new variables though, without having to specify a flag. That would make it easier to use, and also simplify the PR quite a bit. This is similar to how Bats tests are always recognized without any explicit flags.

yetamrra:
OK, I turned on the behavior by default.  That does make the change a lot smaller.  I agree that it seems largely harmless, since people who don't use shflags are pretty unlikely to define functions and variables with those same names.

koalaman:
Great, thanks!

PR #1616, opened 2019-06-18
Add .dockerignore

blueyed:
This explicitly defines included/copied files, to reduce the context
being sent to the Docker daemon initially.

Feel free to close this without merging - it is probably not as relevant as with e.g. Python projects where you would have `.venv` etc, and adds some maintenance overhead when using/copying new files.


koalaman:
Looks good to me, thanks!

PR #1634, opened 2019-07-02
Add portage bash dialect

yetamrra:
Gentoo's portage package manager is built around bash, but with some specific constraints on the environment where the scripts run and a bunch of additional global data.  We can mostly lint it as bash, but it's helpful to be able to have some behavior differences without passing tons of command-line options.

This adds portage as a new shell that should be initially identical to bash, and then turns off one warning as an example of something that could work differently.  If the general approach is reasonable, we'll send more portage-specific changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - json1 format like --format=json but treats tabs as single characters
 - Recognize FLAGS variables created by the shflags library.
 - Site-specific changes can now be made in Custom.hs for ease of patching
+- Add portage as a bash-like shell for ebuild/eclass files.
 - SC2154: Also warn about unassigned uppercase variables (optional)
 - SC2252: Warn about `[ $a != x ] || [ $a != y ]`, similar to SC2055
 - SC2251: Inform about ineffectual ! in front of commands
--- a/shellcheck.1.md
+++ b/shellcheck.1.md
@@ -87,8 +87,9 @@ not warn at all, as `ksh` supports decimals in arithmetic contexts.
 
 :   Specify Bourne shell dialect. Valid values are *sh*, *bash*, *dash* and *ksh*.
     The default is to deduce the shell from the file's `shell` directive,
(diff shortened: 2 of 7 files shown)


koalaman:
I'm not convinced by the use case. If the goal is to suppress frequently ignored warnings, I think a `.shellcheckrc` would be the way to go.

yetamrra:
Yeah, I sent that one because it's small, but then I realized suppressing a warning isn't the most convincing use case :)

We have some other ideas we'd like to send that can't easily be achieved by suppressing warnings, such as: adding a list of portage's built-in global variables; implementing some basic knowledge of the "inherit" command similar to the shflags DEFINE_nnn stuff; tweaking some of the error messages to refer to portage-specific solutions; etc.

We could also potentially just leave the shell as bash and trigger an extra "runningInsidePortage" bool when the file extension is .ebuild or .eclass.  That would make this initial diff much smaller, but leads to some of the checks looking at two conditions instead of one.  Other ideas?

yetamrra:
I updated this to treat portage as bash+flag instead of a separate shell and added a couple of examples of how it could be used.  Please take another look.

XVilka:
Indeed, would be nice to have.

nihilus:
Maybe a 'portagelint' is more appropriate since the constraints is not easy to untangle as a specific dialect etc. Personally I'd suggest to use a wrapper around 'bash -x' for debugging.

chutz:
I am not fluent enough in Haskell to comment on the contents of the patch, but having shellcheck support for portage would certainly be nice. Portage is essentially a dialect of bash, with a lot of extras added on top.

koalaman:
Today I think that a lot of this functionality could more easily be accomplished through something like

```sh
# shellcheck source=portage.bash
true || . /dev/null
```

and have `portage.bash` be a dummy file exporting variables with or without spaces. This would result in the same warnings/suppressions/spellcheck and be useful for any framework without explicit upstream support.

PR #1639, opened 2019-07-12
Fix syntax and indentation in shellcheck.1.md

shak-mar:
Out of interest, I ran the command

    pandoc -s -f markdown-smart -t man shellcheck.1.md -o shellcheck.1

locally, but that produces warnings (previous to this commit).  Checking
the generated manpage, I found the diff to be rendered very badly.
(Broken at terminal width like a normal paragraph).  This commit fixes
the problem.


koalaman:
Thanks!

PR #1654, opened 2019-07-25
Add MacOS link to pre-compiled

Luizm:


koalaman:
This should really be pointing to a stable version rather than the latest commit, but there's no stable version for macOS quite yet

PR #1713, opened 2019-10-02
Running quicktest and quickrun through stack

iboss-ptk:
executing `runghc` and `ghci` in those scripts without stack causing problems for me, so I use stack to execute those commands instead of running it from local bin.


koalaman:
For me it's the other way around, so I'll leave it as is :P

PR #1714, opened 2019-10-02
Allow `read -t 0` to not require -r flag

### LINKED ISSUE #1709: read -t 0 does not require -r
#### For bugs
- Rule Id SC2162
- My shellcheck version (`shellcheck --version` or "online"): 0.4.7 and online

```sh

#!/bin/bash
read -t 0 && IFS='' read -d '' -r std_in

```

^-- SC2162: read without -r will mangle backslashes.

no report, it is ok

iboss-ptk:
Resolve #1709

This PR checks if the command is exactly `read -t 0`, if so then just ignore `-r` requirement for `read` command since it has specific purpose for checking if there is any input to be read.

If used for other means, eg. piping file directly to the command, `-r` is still required since it's not being used as an input detector anymore.


koalaman:
Thanks! I additionally tweaked it to allow things like read -u 3 -t 0

Issue #1724, opened 2019-10-19
TODO: bash: missing support for 'builtin' keyword


balanco:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2154
- My shellcheck version (`shellcheck --version` or 'online'): online
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/bash

builtin echo hello | while builtin read -r ze_variable; do
   echo "${ze_variable}"
done
   
```

#### Here's what shellcheck currently says:
```
Line 4	SC2154: ze_variable is referenced but not assigned.
```
#### Here's what I wanted or expected to see:
no such error message

e-kwsm:
fixed by #1724.

PR #1745, opened 2019-11-01
Fix issue #1724 (bash: missing support for 'builtin' keyword)

jabberabbe:
Fix issue #1724 
Now shellcheck recognises read/written variables in builtins called with the 'builtin' keyword. The parser looks at the first argument passed to `builtin` to determine the type of the other arguments so `builtin cmd ...` is parsed in the same way as `cmd ...` and it automatically parses assignments correctly if the builtin is `export`, `declare`, etc.

With the second commit command-specific checks in `ShellCheck.Checks.Command` are applied to commands called with `builtin` in the same vein as the parser.

It might be useful to add a new warning for unknown builtins because as of now `builtin my-strange-cmd-which-isnt-a-builtin` doesn't trigger any. @koalaman if you agree with me, leave a reply on this message and I will add another commit on this pull request.

---
### Testing
`stack test`:

```
[...]
ShellCheck> Test suite test-shellcheck passed
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## v0.7.1 - soon
 ### Fixed
 - `-f diff` no longer claims that it found more issues when it didn't
-- SC2154 triggers for builtins called with `builtin`
+- SC2154 and all command-specific checks now trigger for builtins
+  called with `builtin`
 
 ### Added
 - SC2254: Suggest quoting expansions in case statements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v0.7.1 - soon
 ### Fixed
 - `-f diff` no longer claims that it found more issues when it didn't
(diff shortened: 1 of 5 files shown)


koalaman:
Nice! How about leveraging `getCommandNameAndToken` from ASTLib.hs instead? It handles the `builtin` case as well as `busybox`

I'm not quite sure about the parser changes. Bash and Ksh both fail to parse `builtin export foo=(bar)` for the same reason, so chances are this should remain the way it is.

jabberabbe:
You are right - `export foo=(bar)` parses correctly while `builtin export foo=(bar)` doesn't. But without changing the parser `builtin export foo=bar` doesn't parse correctly because the parser doesn't see an assignment. What should I do?

koalaman:
I see what you mean. That is indeed annoying. I don't have any good workarounds, so I merged it as-is. Thanks!

PR #1749, opened 2019-11-09
Make image build process a bit simpler

lvjp:
Take full leverage of multi-stage docker build.


koalaman:
Nice, that makes things easier! I'll just merge it and let Travis decides if it works :P

PR #1782, opened 2019-12-25
Add support for pre-commit

ArturKlauser:
[pre-commit](https://pre-commit.com/) is infrastructure to easily set up running any set of tools as git pre-commits on a repo. A tool only needs to provide a hook description (.pre-commit-hook.yaml) in its repo to become usable by the pre-commit infrastructure.

Here I'm adding this hook description, which makes shellcheck easily usable by pre-commit (see added example in README.md). I'm also turning pre-commit loose on this repo and configure it to run automatically on the Github Actions CI pipeline on every push and pull request.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit
+    rev: v2.1.1
+    hooks:
+    -   id: validate_manifest
+
+-   repo: local
+    hooks:
+    -   id: shellcheck
+        name: shellcheck
+        language: docker_image
+        entry: koalaman/shellcheck:stable
+        types: [shell]
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,17 @@ before_install: |
   test "$TRAVIS_BRANCH" = master && TAGS="$TAGS latest" || true
   test -n "$TRAVIS_TAG"          && TAGS="$TAGS stable $TRAVIS_TAG" || true
   echo "Tags are $TAGS"
+  set -e
+  # The version encoded in pre-commit and built version must be in sync. If not,
+  # the pre-commit version was forgotten to be updated.
+  if [ -n "$TRAVIS_TAG" ]; then
+    precommit_tag="$(awk -F: '/^ *entry:/ {print $NF}' .pre-commit-hooks.yaml)"
+    if [ "$TRAVIS_TAG" != "$precommit_tag" ]; then
+      echo "Wrong version number $precommit_tag in .pre-commit-hooks.yaml"
+      sleep 1  # give Travis time to grab the error message
+      exit 1
+    fi
+  fi
 
 # This is in global context and runs for every stage that doesn't override it.
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,27 @@ jobs:
         - source ./.multi_arch_docker
         - set -ex; multi_arch_docker::main; set +x
 
+    # Run pre-commit checks with the shellcheck docker package that was just
+    # deployed.
+    - stage: Pre-commit checks
+      language: python
+      python: 3.7
+      cache:
+        - pip
+        - directories: $HOME/.cache/pre-commit
+      install: pip install pre-commit
+      script: |
+        # Run pre-commit. It also uses shellcheck on itself from the just built
+        # and deployed docker image.
+        # Use the first build tag, or if there isn't any use 'latest'.
+        set -e
+        tag="$(echo "${TAGS/ */}")"
+        tag="${tag:-latest}"
+        echo "tag: $tag"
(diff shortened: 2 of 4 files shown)


koalaman:
Nifty!

I'm a bit skeptical about adding these Github Actions CI test though. This adds a separate set of CI tests on a separate CI platform that only checks ShellCheck's utility scripts with an older version of itself. It seems like a simpler, better and more maintainable idea to have Travis check the scripts with the release itself

ArturKlauser:
Moving it to Travis sounds reasonable enough to reduce maintenance effort.
I'll work on that and let you know when it's ready for another review.

ArturKlauser:
The pre-commit run has now been moved from Github Actions CI to Travis CI.
Please take another look.

IvanBoyko:
@koalaman , @ArturKlauser , this is a great idea!
Maybe you can finish work, so we can use official git repo rather than 3rd party?

ArturKlauser:
Rebased PR to master - applies cleanly at the moment.

IvanBoyko:
@koalaman  , go for it, please :)
Travis build fails for the same reason, as master build, so it's not related to this PR.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit
+    rev: v1.20.0
```
rca0:
```suggestion
    rev: v2.1.1
```

ArturKlauser: Good point, I haven't kept track of pre-commit versions having advanced. I'll fold that into my next push which should hopefully also fix the Travis CI issues in shellcheck HEAD.

ArturKlauser:
rebased to v0.7.1

itxx00:
:+1:

README.md
```diff
@@ -211,6 +211,13 @@ docker run --rm -v "$PWD:/mnt" koalaman/shellcheck:stable myscript
 
 or use `koalaman/shellcheck-alpine` if you want a larger Alpine Linux based image to extend. It works exactly like a regular Alpine image, but has shellcheck preinstalled.
 
+From [pre-commit](https://pre-commit.com/), add this to your `.pre-commit-config.yaml`:
+
+    -   repo: https://github.com/koalaman/shellcheck
+        rev: ''  # Use the sha / tag you want to point at
```
koalaman, line 217: What happens if this is not filled in?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+    rev: v2.1.1
+    hooks:
+    -   id: validate_manifest
+
+-   repo: local
+    hooks:
+    -   id: shellcheck
+        name: shellcheck
+        language: docker_image
+        entry: koalaman/shellcheck:stable
```
koalaman, line 12: I'm not sure how this fits together. Why are there three different versions being specified in the various files?

.travis.yml
```diff
@@ -57,6 +57,27 @@ jobs:
...
+      language: python
+      python: 3.7
+      cache:
+        - pip
+        - directories: $HOME/.cache/pre-commit
+      install: pip install pre-commit
+      script: |
+        # Run pre-commit. It also uses shellcheck on itself from the just built
+        # and deployed docker image.
+        # Use the first build tag, or if there isn't any use 'latest'.
```
koalaman, line 72: Always using `latest` should be fine

.travis.yml
```diff
@@ -65,6 +86,17 @@ before_install: |
   test "$TRAVIS_BRANCH" = master && TAGS="$TAGS latest" || true
   test -n "$TRAVIS_TAG"          && TAGS="$TAGS stable $TRAVIS_TAG" || true
   echo "Tags are $TAGS"
+  set -e
+  # The version encoded in pre-commit and built version must be in sync. If not,
+  # the pre-commit version was forgotten to be updated.
```
koalaman, line 91: This should be in `test/check_release` instead so it can be run before and not after a release

mtekel:
Would be nice to have. Looks like there are already two other repos that wrap shellcheck for pre-commit:

https://github.com/shellcheck-py/shellcheck-py
https://github.com/detailyang/pre-commit-shell

Direct support would be nicer.

dudicoco:
For anyone who's interested, I was able to easily add a pre-commit docker image hook for shellcheck:
```
- repo: local
  hooks:
  - id: shellcheck
    name: shellcheck
    language: docker_image
    entry: koalaman/shellcheck:v0.7.1
    types:
    - shell
```

roberth:
If you use Nix, [pre-commit-hooks.nix](https://github.com/cachix/pre-commit-hooks.nix) supports shellcheck.

```
    hooks.shellcheck.enable = true;
```

koalaman:
After evaluating all the options, the best approach appeared to be a hook in a separate repo auto-generated from Docker tags. This has been added to the README.

The primary benefit of this is that pre-commit stays out of the critical path of ShellCheck releases, like using `docker_image` directly, but additionally allows `pre-commit autoupdate` to work.

PR #1783, opened 2019-12-28
Fix OSX build on Travis

ArturKlauser:
Symlink was failing due to pre-existing target file:

    sudo ln -s /usr/local/bin/gsha512sum /usr/local/bin/sha512sum
    ln: /usr/local/bin/sha512sum: File exists

Build is passing now https://travis-ci.org/ArturKlauser/shellcheck/builds/630262030

PR #1784, opened 2019-12-28
Fix Travis warnings

ArturKlauser:
Fixing the following Travis build config validation warnings:
  * root: deprecated key sudo (The key `sudo` has no effect anymore.)
  * root: missing os, using the default linux
  * deploy: key local-dir is not underscored, using local_dir
  * language: value sh is an alias for shell, using shell
  * root: key matrix is an alias for jobs, using jobs

The linux builds pass https://travis-ci.org/ArturKlauser/shellcheck/builds/630319022
the osx build is broken for a different reason #1783.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,17 @@
-
-sudo: required
-
-language: sh
+language: shell
+os: linux
 
 services:
   - docker
 
-matrix:
+jobs:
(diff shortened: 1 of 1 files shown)


koalaman:
Merging and hoping for the best. Thanks!

PR #1785, opened 2020-01-02
Add multi-architecture Docker image build

ArturKlauser:
This builds the docker images as multi-architecture images, which is a recent feature supported by docker's buildx extension. The corresponding Dockerfile simply reuses the shellcheck binaries that are already built and deployed for multiple architectures. See how my [test images](https://hub.docker.com/repository/docker/arturklauser/shellcheck-alpine/tags?page=1) report availability for multiple architectures. Each client pulling from Docker Hub automatically pulls only the appropriate architecture version for itself, so there is no additional overhead for the client. Clients don't need experimental features enabled (only the builder need them).

The docker build and deployment is hooked into the existing Travis CI/CD pipeline, conditioned on all other builds and tests passing. You can see a [sample run](https://travis-ci.org/ArturKlauser/shellcheck/builds/631826071) on my fork.

### DIFF
--- a/.compile_binaries
+++ b/.compile_binaries
@@ -19,12 +19,6 @@ build_linux() {
   do 
     cp "shellcheck" "deploy/shellcheck-$tag.linux-x86_64"; 
   done
-
-  # Linux Alpine based Docker image
-  name="$DOCKER_BASE-alpine"
-  DOCKER_BUILDS="$DOCKER_BUILDS $name"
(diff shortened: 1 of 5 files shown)


koalaman:
Oops, I put off reviewing this for too long and forgot. Seems to work and do the right thing, thanks! I'll test more thoroughly soon

PR #1802, opened 2020-01-21
SC2016, repair false error for M language parser

szydell:
mumps is a compiler and parser for the M language from the gt.m.
(https://sourceforge.net/projects/fis-gtm/)


koalaman:
A good alternative is often to just blacklist the whole command to make the behavior more predictable and less finicky. Since I don't know how the command is used, I'll just take your word for it ^^

szydell:
Thanks!

PR #1814, opened 2020-01-31
Don't trigger SC2154 (unassigned var) in `-n`/`-z` expressions #1583

donnerpeter:


koalaman:
I totally dropped the ball here, but this is now merged with some additional improvements in 50067ddf94 to handle cases like `x=$(var); if [ -z "$x" ]; then ..` that would otherwise emit an unused warning. Thanks!

PR #1816, opened 2020-02-02
Various cleanups and refactorings

josephcsible:
Explanations for each change are in their corresponding commit messages.


koalaman:
Thanks! ShellCheck was my first Haskell project and that's especially obvious in these older files :P

PR #1822, opened 2020-02-05
SC2191: Tighten index checks

yetamrra:
When adding a value containing an equals sign to an indexed array, the
left side is treated as an index if it looks like [N]=val and N is
numeric.  SC2191 currently warns about anything that looks like key=val
even though non-numeric values of key will never be treated as an index.
This causes spurious warnings for the common pattern of building up
program arguments in an array, such as:
  args=(
    --dry-run
    --in="${my_var}"
    --out=/some/path
    -f
  )
  /bin/program "${args[@]}"

Since only numeric expressions can be a valid index for an indexed
array, only emit SC2191 if the left side of a literal string containing
an equals looks numeric.  Other more complicated constructs should still
warn because shellcheck doesn't know if they may evaluate to a numeric
result.  Associative arrays still warn because a non-numeric left side
is a valid subscript.


koalaman:
This fixed a whole lot of false positives. Thanks!

PR #1825, opened 2020-02-09
Use findM instead of filterM

josephcsible:
Using filterM makes us run the monadic predicate on every list element.
Use findM instead so that we can stop once we find a matching one.


koalaman:
Very nice catch!

PR #1826, opened 2020-02-09
Use the Identity monad to avoid unnecessary uses of fromJust

josephcsible:


src/ShellCheck/Analytics.hs
```diff
@@ -2595,7 +2593,7 @@ checkTildeInPath _ (T_SimpleCommand _ vars _) =
             warn id 2147 "Literal tilde in PATH works poorly across programs."
     checkVar _ = return ()
 
-    hasTilde t = fromMaybe False (liftM2 elem (return '~') (getLiteralStringExt (const $ return "") t))
+    hasTilde t = '~' `elem` onlyLiteralString t
```
koalaman: Haha, I don't even know what to say about the previous version of this

koalaman:
Much better, thanks!

PR #1857, opened 2020-03-03
Add conda install instructions

lukelbd:
This adds installation instructions for anaconda users. The fact that `shellcheck` is [available on anaconda](https://anaconda.org/conda-forge/shellcheck) is very, very useful for users that work on remote servers / compute clusters without administrative privileges.


koalaman:
Thanks!

PR #1866, opened 2020-03-10
Revert one incorrect change from 172aa7c

josephcsible:


josephcsible:
Bizarrely, I can't reproduce the bug that I expect this to be fixing.

koalaman:
Nice catch! This fixed a bug in 2014 when ShellCheck would parse `foo{1,2,3}` as a single literal because it didn't know to stop reading unquoted string literals at `{`s.

That's issue's long since been fixed, so this test no longer serves a purpose. Even if a brace expansion did fail to parse and was treated as a literal, it would result in two literals that would fail the pattern match this check was behind.

I fixed up and modernized this check in a57f6d2886e899133218e238fb1a3b26f948010d, and just dropped the condition as part of that. Thanks!

PR #1872, opened 2020-03-15
Simplify checkForInQuoted

josephcsible:
* Avoid some unnecessary fmaps
* Reuse an identical pattern-match for two guards
* Apply De Morgan's law
* Use forM_ to avoid an unnecessary where


koalaman:
Thanks!

PR #1873, opened 2020-03-16
Simplify checkWhileReadPitfalls

josephcsible:
* Clean up usage of not
* Use a case match instead of sequence_ and a do block


koalaman:
Thanks!

PR #1876, opened 2020-03-16
recognize `: ${parameter=word}` as assignment

### LINKED ISSUE #1758: ${parameter=word} not recognized as assignment
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2223

`${parameter=word}` is very similar `${parameter:=word}`, except that it substitutes parameter even if it is defined *null*. Both are [specified by POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02). Shellcheck recognizes the latter as an assignment, but not the former and issues a confusing warning.

```sh
#!/bin/sh

: "${foo:=foo}"  # works
: "${bar=bar}"  # warns "SC2154: bar is referenced but not assigned."

echo "$foo $bar"
```

> SC2154: bar is referenced but not assigned.

nothing

girst:
This fixes #1758 and adds a test for it.

for reference, commit 4fc48998037fd52c07c29b62cf25584ba738cead introduced the `:=` variant, and 3c5c74ff047ba99451b3a1a7689f5c1b84d1634b adds (broken) support for both variants (but misses this place).


koalaman:
Thanks!

PR #1880, opened 2020-03-20
Mark that base >= 4.8.0.0 is required

josephcsible:
We've actually already required base >= 4.8.0.0 since commit a8376a0 (in which we first used `<$>` without an import, which wasn't in the Prelude prior to this version). Since then, we've also made use of other more substantial features that de-facto require base >= 4.8.0.0 since they require GHC 7.10, such as `DeriveAnyClass`.

### DIFF
--- a/ShellCheck.cabal
+++ b/ShellCheck.cabal
@@ -45,9 +45,7 @@ library
     build-depends:
       aeson,
       array,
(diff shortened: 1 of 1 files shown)


koalaman:
Good to make this explicit. Thanks!

PR #1893, opened 2020-03-28
Fix #1892: Use pattern synonyms to clean up AST

### LINKED ISSUE #1892: Do something with the monstrosity that is delve
Bugs like #1858 only happen because we handwrite `delve`: https://github.com/koalaman/shellcheck/blob/7963eeab9d51b5210f21de13eaf2140805287cef/src/ShellCheck/AST.hs#L188-L287

That implementation is super error-prone (as that bug illustrated). We should decide on a refactoring that avoids having to write it, since it's basically just a monomorphic `traverse`.

Commit 0bff853 is a proof-of-concept that it can be replaced with generated code. I didn't just PR that because I imagine you'll frown upon that particular implementation due to its use of Template Haskell and bringing in another dependency. If you agree with the idea, though, I can look into other approaches to do it, such as generics.

josephcsible:


koalaman:
This is awesome!

I tested it on multiple distros and versions to ensure it was portable, and ran it on thousands of files to make sure there were no surprises, and it worked perfectly. Not only did it get rid of some boilerplate and a dirty hack, it also improved total runtime by ~10%!

Great job, thanks!

josephcsible:
Side note: how do you do performance testing/measure runtime?

PR #1900, opened 2020-04-05
Clean up AnalyzerLib

josephcsible:


koalaman:
Thanks!

src/ShellCheck/AnalyzerLib.hs
```diff
@@ -354,8 +354,8 @@ getClosestCommand tree t =
 
 -- Like above, if koala_man knew Haskell when starting this project.
 getClosestCommandM t = do
-    tree <- asks parentMap
-    return $ getClosestCommand tree t
+    params <- ask
+    return $ getClosestCommand (parentMap params) t
```
koalaman, line 358: Is there a particular reason why this is better?

josephcsible: `asks f` is equivalent to `fmap f ask`, and fmaps of something that's immediately being fed into a monadic bind are always unnecessary, and slower in some monads.

PR #1901, opened 2020-04-05
Mostly get rid of bracedString

josephcsible:
`bracedString` is a dangerous partial function, and it's simple enough to replace it with its safe equivalent everywhere possible. Do that everywhere except for `isDefaultAssignment` in `checkSpacefulness`, where I don't see a way to prove that it's actually always safe. Also, move its definition there so that no new uses of it appear.

### DIFF
--- a/src/ShellCheck/ASTLib.hs
+++ b/src/ShellCheck/ASTLib.hs
@@ -134,10 +134,6 @@ isUnquotedFlag token = fromMaybe False $ do
     str <- getLeadingUnquotedString token
     return $ "-" `isPrefixOf` str
 
--- Given a T_DollarBraced, return a simplified version of the string contents.
-bracedString (T_DollarBraced _ _ l) = concat $ oversimplify l
-bracedString _ = error "Internal shellcheck error, please report! (bracedString on non-variable)"
-
 -- Is this an expansion of multiple items of an array?
 isArrayExpansion (T_DollarBraced _ _ l) =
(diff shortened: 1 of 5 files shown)


koalaman:
`oversimplify`, née `deadSimple`, is a relic from the beginning of time when it seemed useful to recognize commands by a canonical string representation, e.g. `case oversimplify cmd of "grep":"${VAR}":_ -> ...` (to suggest `-e`), thereby matching `grep $foo` and `grep "$(bar)"` at the same time.

This turned out to be a bad idea because 1. it's not actually that convenient, and e.g. fails for `grep $x$y` (`["grep", "${VAR}${VAR}"]`). 2. it's not flexible enough, since it discards information about quoting status and similar, 3. it discards information about the tokens for error reporting.

For the most part, I've tried to replace `oversimplify` with `getLiteralString` & friends. I was initially skeptical of this PR until I realized that `bracedString` itself was still implemented via `oversimplify`.

Do you have any suggestions for how we could aim to get rid of both problems?

PS: Thanks for all your contributions. Please let me know if there's anything I can do in return (e.g. a mention somewhere, or at referral to one of Facebook's FP teams)

josephcsible:
I can't think of any good way to do it all at once, since `oversimplify` does so much "magic", and `bracedString` is such a thin wrapper around it.  I think we'll be stuck replacing each individual use as one-offs.

PR #1918, opened 2020-04-12
Clean up and optimize getSuspiciousRegexWildcard

josephcsible:


koalaman:
Thanks for all the PRs!

dimo414:
It looks like this or one of the earlier PRs broke some builds: https://travis-ci.org/github/koalaman/shellcheck/builds/674194723

josephcsible:
@dimo414 Looks like it was actually caused by 64c31d9, as `fromRight` didn't exist in GHC 8.0 or older. If we still want to support versions that old, then just that commit should be reverted.

josephcsible:
#1925

PR #1925, opened 2020-04-25
Revert "Use fromRight instead of reimplementing it"

josephcsible:
We still support GHC 8.0, which didn't have fromRight.

This reverts commit 64c31d91422e3c48cc7ed0176b069fa11456cee8.


koalaman:
Thanks!

PR #2076, opened 2020-10-21
Add Mega-Linter in list of including platforms

nvuillam:
If you have an icon or a banner, please comment and i'll add it on your linter page :)

https://github.com/nvuillam/mega-linter/blob/master/docs/descriptors/bash_shellcheck.md#readme


koalaman:
Thanks, but this list is not meant to be exhaustive and only features a few of the most common platforms. Good luck, and bump when it takes off ^^

Issue #1336, opened 2018-09-11
readonly not being flagged as part of SC2155


jgrobbel:
#### For bugs
- Rule Id (if any, e.g. SC1000): 
- My shellcheck version (`shellcheck --version` or "online"): 
online and 
shellcheck --version
ShellCheck - shell script analysis tool
version: 0.5.0
- [x] I read the issue's wiki page, e.g. https://github.com/koalaman/shellcheck/wiki/SC2155
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```
#!/usr/bin/env bash

set -o pipefail -o errtrace -o errexit -o nounset
trap 'echo >&2 trap on error \(rc=${PIPESTATUS[@]}\) near line $LINENO' ERR


test_func() {
  local timeout
  readonly timeout="$(./script_rc_200.sh)"
  echo "Should not see this"
}

test_func
```

 Output:
```
 ./readonly.sh
trap on error (rc=200) near line 9
Should not see this
```

#### Here's what shellcheck currently says:

$ shellcheck myscript
No issues detected!

#### Here's what I wanted or expected to see:

$ shellcheck ./readonly.sh

In ./readonly.sh line 9:
  readonly timeout="$(./script_rc_200.sh)"
         ^-- SC2155: Declare and assign separately to avoid masking return values.

koalaman:
There's currently no warning because you can't split this into `readonly timeout; timeout=foo` like with the others, but there should arguably be a separate warning explicitly suggesting declare+assign+readonly, e.g. `local timeout; timeout="foo"; readonly timeout`.

rpdelaney:
See also #731

keith:
I'm not sure if there was more of a reason this wasn't fixed over the past few years but I submitted one here: https://github.com/koalaman/shellcheck/pull/2077

Note that both of these work as a workaround for this issue:

```sh
foo=$(exit 1)
readonly foo
```

```sh
foo=$(exit 1)
declare -r foo
```

Even if there isn't a workaround as a user I would prefer to get hit by this warning and have to explicitly disable it than be surprised by a runtime failure

PR #2077, opened 2020-10-24
Add readonly to SC2155

keith:
This adds a warning for readonly masking the return value of function.
This is mentioned in the wiki
https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-readonly
but didn't actually produce a warning.

Fixes https://github.com/koalaman/shellcheck/issues/1336


koalaman:
Thanks!

PR #2097, opened 2020-12-01
Add MacPorts as installation option in README.md

ylluminarious:
Thought it was worth mentioning MacPorts as an installation option on macOS.


koalaman:
Thanks!

PR #2119, opened 2020-12-28
Various refactorings

josephcsible:


src/ShellCheck/Analytics.hs
```diff
@@ -3858,13 +3858,13 @@ groupByLink :: (a -> a -> Bool) -> [a] -> [[a]]
 groupByLink f list =
     case list of
         [] -> []
-        (x:xs) -> g x [] xs
+        (x:xs) -> foldr c n xs x []
   where
-    g current span (next:rest) =
+    c next rest current span =
```
Tylerwbrown, line 3863: Why was this changed from `g` to `c`?

Tylerwbrown (approved):
Pretty much all these changes are an improvement in my opinion, for what it's worth.

src/ShellCheck/Analytics.hs
josephcsible: Before, `g` handled both the cons and nil cases, since it was doing explicit recursion. Now that `foldr` is handling the recursion, there's two different functions, so I called them `c` for the cons case and `n` for the nil case.

koalaman:
Thanks!

shellcheck.hs
```diff
@@ -507,7 +507,7 @@ ioInterface options files = do
       where
         find filename deflt = do
             sources <- findM ((allowable inputs) `andM` doesFileExist) $
-                        (adjustPath filename):(map (</> filename) $ map adjustPath $ sourcePathFlag ++ sourcePathAnnotation)
```
demkaid, line 510: - [ ] 

- [ ] `[](url)`

Issue #2057, opened 2020-09-17
SC2016 false positive (podman)


adrianreber:
This is the same as https://github.com/koalaman/shellcheck/issues/1161 just for Podman.

I am running one container with docker and one with podman. The same command-line gives an error when using podman.

This gives no SC2016:
```
docker run --read-only -d alpine /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'
```
This gives a SC2016:
```
podman run --read-only -d alpine /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'
```

adding `,"podman" -- like above` to src/ShellCheck/Analytics.hs should fix this.

cornfeedhobo:
I have another failing example:

```plain
        local msg="${2:-Command '$1' does not exist!}"
                                ^--^ SC2016: Expressions don't expand in single quotes, use double quotes for that.
```

However, works as expected:
```plain
Command 'flimflam' does not exist!
```

cornfeedhobo:
I have another failing example:

```plain
        local msg="${2:-Command '$1' does not exist!}"
                                ^--^ SC2016: Expressions don't expand in single quotes, use double quotes for that.
```

However, works as expected:
```plain
Command 'fake-command' does not exist!
```

kolyshkin:
@cornfeedhobo I think that your issue is not related -- please file it separately.

cornfeedhobo:
@kolyshkin Okay. Made one at #2168

PR #2134, opened 2021-01-28
Whitelist podman for SC2016 about '$var'

kolyshkin:
Same as 08d2eef4111afde4 but for podman.

Fixes https://github.com/koalaman/shellcheck/issues/2057


koalaman:
Thanks!

Issue #1988, opened 2020-06-15
#!/bin/busybox sh isn't recognized as a valid shebang


austin987:
#### For bugs
- Rule Id: SC1008
- My shellcheck version : online 
- [X] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2086)
- [X] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### For new checks and feature suggestions
- [X] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/busybox sh
true

```

#### Here's what shellcheck currently says:
```
$ shellcheck myscript
 
Line 1:
#!/bin/busybox sh
^-- SC1008: This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.
```

#### Here's what I wanted or expected to see:
Nothing.

I noticed this in a real world script: https://github.com/SolidHal/PrawnOS/blob/master/resources/BuildResources/initramfs-init

the script itself is a posix sh script, invoked via busybox. I'm not sure how common it is to call busybox directly, but given that it's a shell script, perhaps shellcheck should accept that shebang.

austin987:
See also https://github.com/koalaman/shellcheck/issues/853, which concerns busybox's ash shell (noticed while I set shell=sh for this script)

grische:
@austin987 This seems to be supported as of https://github.com/koalaman/shellcheck/pull/2139

PR #2139, opened 2021-02-06
add support for `/bin/busybox sh` shebang

austin987:
For #1988

First time with Haskell, so please forgive any obvious mistakes (happy to correct with guidance) ;).

In particular, not totally sure about the prop_checkShebang bits, I tried to follow earlier conventions there.


koalaman:
Thanks! Shell detection code had ended up duplicated for no good reason, so I merged this and other accumulated improvements into a single function.

PR #2140, opened 2021-02-07
Update simple installer snippet

felipecrs:


README.md
```diff
@@ -259,12 +259,18 @@ On Redhat/Fedora/CentOS, `yum -y install xz`.
...
-scversion="stable" # or "v0.4.7", or "latest"
-wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
-cp "shellcheck-${scversion}/shellcheck" /usr/bin/
+scversion="v0.7.1"
+curl -fsSL "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion}.$(uname -s).$(uname -m).tar.xz" \
+    | sudo tar -xJf - --strip-components=1 -C /usr/local/bin shellcheck-${scversion}/shellcheck
 shellcheck --version
 ```
 
+If you want to install the latest version, you can use something like with `jq`:
```
koalaman, line 268: In this case you can more easily just use scversion="stable"

felipecrs: You are right, but this is to make the process more generic compared to other GitHub projects which do not publish the package as you do.

Anyway, it's your call. What's your preference?

koalaman:
I'll leave the existing instructions for now, since they're relatively simple even if not universally copy-pastable.

PR #2145, opened 2021-02-12
Allow `env` to have flags and variables in shebang

### LINKED ISSUE #2143: SC1008 Shell detection fails on "#!/usr/bin/env -S bash"
#### For bugs
- Rule Id: SC1008
- My shellcheck version: "online" (2021-02-10, master at 15ff87cf80bfcc23fa4a3d0412594e17959dd1b0)

```sh
#!/usr/bin/env -S bash
```

```
Line 1:
#!/usr/bin/env -S bash
^-- SC1008: This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.
```

No output, `bash` is deduced from the shebang just like with `#!/usr/bin/env bash`.

avdv:
The `env` command has a `-S,--split-string` option which enables having arguments for the command in a shebang.

Also, one can use variable assignments for the command since `env` treats only the first word without a `=` character as the
program to run.

Fixes #2143.

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -3190,6 +3190,7 @@ prop_readScript2 = isWarning readScript "#!/bin/bash\r\necho hello world\n"
 prop_readScript3 = isWarning readScript "#!/bin/bash\necho hello\xA0world"
 prop_readScript4 = isWarning readScript "#!/usr/bin/perl\nfoo=("
 prop_readScript5 = isOk readScript "#!/bin/bash\n#This is an empty script\n\n"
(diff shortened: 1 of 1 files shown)


koalaman:
Thanks!

matthewpersicowork:
What OS is that on? I've never seen that.

avdv:
> What OS is that on? I've never seen that.

This is from GNU coreutils' env. See http://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html#env-invocation

So, this applies to Linux, Hurd, probably darwin and windows with cygwin / mingw and so on.

Issue #2132, opened 2021-01-25
Feature Request: Warn about trailing backslashes in comments (similar to SC2215)


schaetzc:
#### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [X] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

In below scripts a part of a multi-line command was commented out. However, the comment also disabled a backslash at the end of the line and therefore broke the multi-line command into two commands by accident.

```sh
#!/bin/bash
sed -i \
  -e s/a/b/ \
  # -e s/x/y/ \
  /path/to/file
```

or

```sh
#!/bin/bash
sed -i -e s/a/b/ # -e s/x/y/ \
/path/to/file
```

#### Here's what shellcheck currently says:
```
No issues detected!
```
#### Here's what I wanted or expected to see:

It would be very nice if shellcheck gave a warning about trailing backslashes in comments. The warning could be similar to [SC2215](https://github.com/koalaman/shellcheck/wiki/SC2215). Alternatively, SC2215 could be extended by these kind of mistakes.

m-ildefons:
After reading https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html (section 2.2.1) and rule #9 on token recognition further down on that page, I belive this is a bug not in shellcheck, but in the shell (bash in this case, but dash has this behaviour too)

Your first snippet should be interpreted by the shell just the same as:
```
#!/bin/bash
sed -i -e s/a/b/
```

Since the <backslash> right before the <newline> inside the comment should be expanded before tokenizing, joining the lines, the line following the comment should be part of the comment in its entirety. The same applies to the second example.

I've made a PR just for fun, but I'm not really sure, that that behavior is what we actually want.

koalaman:
I merged the PR with some changes! To avoid false positives it only warns about lines where the previous line was continued (see #2146), so it catches the first and most common case, but not the second. Great suggestion, thanks!

PR #2146, opened 2021-02-14
Error on backslash in comment

m-ildefons:
- Report error in case of a backspace in a comment

Backspaces in comments are no good. In most cases they are the result of
commenting out a longer line, that was broken down. This usually results
in the shell treating the following lines as their own commands on their
own lines instead of as parts of the longer, broken down line.

There are counter examples, but their usage of the backslash to break the line is rather academic anyways.

The following will result in an error:
```
echo "bla 1" \
  #&& echo "bla 2" \
  && echo "bla 3"
```
The following would be the output for the example above:
```
In backslash.sh line 2:
echo "bla 1" \
^-- SC1073: Couldn't parse this simple command. Fix to allow more checks.


In backslash.sh line 3:
  #&& echo "bla 2" \
                    ^-- SC1072: Backslash in or directly after comment. Fix any mentioned problems and try again.
```

The following would be a (rather academic) example of a false positive:
```
echo "bla 1"; \
  #echo "bla 2"; \
  echo "bla 3";
```

I would prefer to emit a warning and continue in this case, but as far as my very limited understanding of the code goes, the parser can only fail completely or be completely silent. Since all tokens in a comment are consumed by the parser, but then discarded, I see no way to achieve that. Also for some reason, I think this throw the numbering for the `SC1XXX` errors off.
Especially after reading https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html part 2.2.1, I don't think erroring out is the way to go. Maybe you have more insight on this, @koalaman ?

related to #2132

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -1039,9 +1039,14 @@ readComment = do
     unexpecting "shellcheck annotation" readAnnotationPrefix
     readAnyComment
 
+prop_readAnyComment = isOk readAnyComment "# Comment"
+prop_readAnyComment1 = not $ isOk readAnyComment "# Comment \\\n"
 readAnyComment = do
     char '#'
-    many $ noneOf "\r\n"
+    comment <- many $ noneOf "\\\r\n"
+    bs <- many $ oneOf "\\"
+    unless (null bs) (fail "Backslash in or directly after comment")
+    return comment
 
 prop_readNormalWord = isOk readNormalWord "'foo'\"bar\"{1..3}baz$(lol)"
 prop_readNormalWord2 = isOk readNormalWord "foo**(foo)!!!(@@(bar))"


m-ildefons:
I thinnk this is not a failure with my commit:
```
[...]
Error: cabal-install: no bottle available!
[...]
Error: gnu-tar: no bottle available!
[...]
```
How can I retrigger the check?

koalaman:
Sorry for taking way more than my sweet time on this. These bugs hard to spot for humans, easy to spot for ShellCheck, and occurs with decent frequency (~1 in 2,000 scripts), so this is a very helpful warning!

The [DevGuide](https://github.com/koalaman/shellcheck/wiki/DevGuide) has some tips for the implementation. ShellCheck mostly tries to parse according to spec, and instead accumulates a list of likely problems (using `parseProblem[At]`) as part of a successful parse, in addition to the resulting AST. Simply using that instead of `fail` would do what you want.

It appears that there's quite a lot of doc comments with [command examples](https://github.com/facebook/rocksdb/blob/3bfd3ed2f393c8d659822fba0860b9dde22efb0b/tools/regression_test.sh#L29-L37) and [correctly commented out](https://github.com/apache/incubator-mxnet/blob/11cd73f5add64919dcb6ddc3fea5bbb6397314d8/ci/docker/runtime_functions.sh#L257-L262) code though, drowning out the actual bugs.

A higher signal-to-noise ratio heuristic appears to be to only warn if the previous line was continued. This check can be done as part of line continuations handling. This happens in the context in which they appear, so mostly `spacing` since it occurs as part of the space between words (`foo\` followed by a line `bar` would instead by handled by `readNormalLiteral` since this is equivalent to the literal `foobar`). Conveniently, this function also consumes comments so it can check as part of that. If it hadn't, the alternative would be a still very simple (but less efficient) `try . lookAhead $ check here` to inspect the upcoming text.

I added this in 98952df35b8 on top of your branch. Thanks!

PR #2166, opened 2021-03-06
suppress ntlm error messages in Windows build

avoidik:
when building for windows there are many error message like below

```
003a:err:winediag:SECUR32_initNTLMSP ntlm_auth was not found or is outdated. Make sure that ntlm_auth >= 3.0.25 is in your path. Usually, you can find it in the winbind package of your distribution.
```


koalaman:
Thanks!

Issue #2189, opened 2021-03-26
Warning for "check exit code directly" is misleading when checking for non-zero exit code


jyn514:
#### For new checks and feature suggestions
- [x] https://www.shellcheck.net/ (i.e. the latest commit) currently gives a misleading warning about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/sh
false
if [ $? -ne 0 ]; then
        echo 1;
fi
```

#### Here's what shellcheck currently says:

Line 3:
if [ $? -ne 0 ]; then
     ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.

This is very misleading - `if false` will have the *opposite* effect from the current code. I had to run `false; echo $?` a few times to convince myself that I was right and shellcheck was wrong. Shellcheck should only suggest a fix if preserves the same behavior.

#### Here's what I wanted or expected to see:

Line 3:
if [ $? -ne 0 ]; then
     ^-- SC2181: Check exit code directly, not indirectly with $?.

Did you mean: 
if ! false; then

brother:
Basic arithmetic is needed indeed.
I think the [wiki page](https://github.com/koalaman/shellcheck/wiki/SC2181) is fairly clear with examples of these things (but do propose expansion there if you have some, apparently they are needed).

koalaman:
The message will now try to suggest `!` when appropriate. Thanks!

jyn514:
That's awesome! Thank you :D

PR #2201, opened 2021-04-12
Fix warning message for SC2181.

dandugula:
Fix for issue #2189.


koalaman:
I went the extra mile and tried to make it suggest `!` when appropriate in 99f6554c9, since this keeps it more actionable. Thanks!

PR #2211, opened 2021-04-20
Fixed comments that were parsed as haddock errors

lorenzo:
When compiling ShellCheck with `-haddock`, the were some comments
containing invalid haddock syntax, whchi prevented the docs from being
generated in hackage.

I fixed the issue by adding another dash, which signals the parser that
it should not be interpreted as documentation, but just comments.


lorenzo:
After this is merged, a new release  for hackage will fix the missing docs.

juhp:
Ah apparently this got fixed by 9e60b3e

PR #2215, opened 2021-04-21
Enable publishing to ghcr.io (Github container registry)

dbast:
Dockerhub got unfortunately rate limited regarding pulling containers.

By publishing to the Github container registry (can be found here https://github.com/koalaman?tab=packages) that rate limit can be avoided and makes pipelines using shellcheck via Docker more reliable.

Merging this requires adding a token called GITHUB_GHCR_TOKEN (with the capability to push containers) to the repo secrets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,7 @@ jobs:
         env:
           DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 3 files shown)


dbast:
Ran into this several time, but seems like there is not much interest in this... can close the PR in that case.

koalaman:
Sorry, I have very little context for this. Can it simply push the same image instead of building a second one? Seems like that would be faster and easier to maintain.

felipecrs:
@koalaman docker stores cache for the images built. The second built with a different tag will be almost instantaneous.

unguiculus:
BuildKit supports specifying multiple tags. You can build and push to multiple registries with one command.

PR #2234, opened 2021-05-18
move readme to extra-doc-files and add changelog to releases

juhp:
Use extra-doc-files (available in Cabal 1.18 and later) and include the CHANGELOG in releases
(then it will be visible on Hackage, etc)


koalaman:
Thanks!

PR #2267, opened 2021-06-27
Problem: 'then' words in pcs scripts confuse ShellCheck

### LINKED ISSUE #2266: SC1010: 'then' words in Pacemaker-related scripts confuse ShellCheck
#### For bugs
- Rule Id (if any, e.g. SC1000): SC1010
- My shellcheck version (`shellcheck --version` or "online"): 0.7.2 (fresh master)
- [ ] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC1010)

Pacemaker command taken from the official [Red Hat resource](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/high_availability_add-on_reference/s1-orderconstraints-haar) fails

The script to reproduce the issue is as follows:
```sh
#!/bin/bash

pcs constraint order VirtualIP then dummy_resource kind=Optional
```

```
In ./pcs-test.sh line 3:
pcs constraint order VirtualIP then dummy_resource kind=Optional
                               ^-- SC1010: Use semicolon or linefeed before 'then' (or quote to make it literal).

For more information:
  https://www.shellcheck.net/wiki/SC1010 -- Use semicolon or linefeed before ...

```

No warning. Reason: bash seems to process the script with no issue, so it sounds like a parser problem.

typeundefined:
Pacemaker-related scripts often use 'then' word unquoted and it does
work in bash. Example:

```
pcs constraint order VirtualIP then dummy_resource kind=Optional
```

This instruction causes SC1010 warning which seems to be a
false-positive.

It seems like 'then' word itself is not a keyword in bash. I guess that
it is understood as a keyword only in context of `if` clause.

Solution:
Exclude 'then' from the list of terminal tokens in `readNormalWord`
function.

Closes #2266

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -1068,7 +1068,7 @@ prop_readNormalWord9 = isOk readSubshell "(foo\\ ;\nbar)"
 prop_readNormalWord10 = isWarning readNormalWord "\x201Chello\x201D"
 prop_readNormalWord11 = isWarning readNormalWord "\x2018hello\x2019"
 prop_readNormalWord12 = isWarning readNormalWord "hello\x2018"
-readNormalWord = readNormalishWord "" ["do", "done", "then", "fi", "esac"]
+readNormalWord = readNormalishWord "" ["do", "done", "fi", "esac"]
 
 readPatternWord = readNormalishWord "" ["esac"]
 


koalaman:
You're right, `then` is not a keyword in this context, but it also wasn't meant as one so this is a false positive. The behavior is intentional and meant to catch issues like these, which would no longer trigger:

```
if [ $value -eq 0 ] then
if grep -q word file then
```

ShellCheck parses them (and yours) according to spec, but emits the warning. It's overall a fairly helpful warning for newbies, but I definitely see how it'd get tedious in this case. Hopefully you're ok helping out the newbies by instead silencing your own instances permanently with a `disable=SC1010` in a `.shellcheckrc` in the project root or your home dir.

Issue #2286, opened 2021-07-19
Bug: ./quickrun does not appear to work anymore


zachriggle:
Hello again @koalaman and thanks for your excellent work on ShellCheck.

Overall this post is about what I'm doing to build ShellCheck, the problems I've encountered (everything builds, runs, and installs -- but it's slow) and looking for support with the `./quickrun` script.

## Background

I'm attempting to make local modifications to support Zsh for my personal use, and I am very new to Haskell and Cabal (and have still not used Stack).  I understand this is something that you don't wish to support this -- as `zsh` support was removed several years ago -- but it's a fun exercise for me in learning Haskell and hacking on a side project.

## Building ShellCheck

I have successfully built via `cabal install` and `cabal build`.  Thanks for the clear instructions in the README.

However, the process of building ShellCheck via `cabal install` is very slow (despite using almost no CPU and very little RAM).  I am looking for a way to cut down on the edit-compile-run cycle for testing new changes.

I did discover that I can speed up the process a bit by instead running `cabal build -j12 -O0` which skips some packaging steps, however the build is still done sequentially (one file at a time, despite `-j12`) and takes quite a bit of time.

Can you offer any tips for speeding up the build?  `build` seems faster than `install`, which makes sense -- but I've got 12 cores and 32GB of RAM and ...

### cabal install

For a fresh build (i.e. I removed the `dist-newstyle` directory), `cabal install` takes 3 minutes.  Note that "99%" CPU shows that a single CPU core was used the entire time, despite `-j12` should use all of them.

```
$ time cabal install -j12 -O0
Wrote tarball sdist to
/Users/zachriggle/github.com/shellcheck/dist-newstyle/sdist/ShellCheck-0.7.2.tar.gz
Resolving dependencies...
Build profile: -w ghc-8.10.5 -O0
In order, the following will be built (use -v for more details):
 - ShellCheck-0.7.2 (lib) (requires build)
 - ShellCheck-0.7.2 (exe:shellcheck) (requires build)
Starting     ShellCheck-0.7.2 (lib)
Building     ShellCheck-0.7.2 (lib)
Installing   ShellCheck-0.7.2 (lib)
Completed    ShellCheck-0.7.2 (lib)
Starting     ShellCheck-0.7.2 (exe:shellcheck)
Building     ShellCheck-0.7.2 (exe:shellcheck)
Installing   ShellCheck-0.7.2 (exe:shellcheck)
[...]

PR #2288, opened 2021-07-30
bats: Add check for useless negation (SC2293)

### LINKED ISSUE #2273: bats: Check for negation that won't fail the test

See https://github.com/bats-core/bats-core/issues/462 for more context.
In short: Using negation on a statement (that is not the last) does not trigger a test failure for bats because bash does not trigger the error trap on that. Since this is a deliberate bash decision, we need to educate users about it.

```bash
#!/usr/bin/env bats

@test "name" {
    ! true # this line should fail the test but doesn't
    true
}
```

Nothing

Warning on the line `! true`: "Negation won't trigger test failure due to bash's design decision not to trigger the error trap on negated statements."

martin-schulze-vireso:
Fixes #2273 

This is the first time I have used Haskell for something useful so please bear with my beginner mistakes. I could not get `./quicktest` to tell me whats wrong with my checks, so some help with this would be appreciated.

Also note that this should remain a draft until [bats 1.5.0 has actually been released](https://github.com/bats-core/bats-core/issues/475).

Please also review the companion wiki pages:

- https://github.com/koalaman/shellcheck/wiki/SC2314
- https://github.com/koalaman/shellcheck/wiki/SC2315

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -4853,19 +4853,30 @@ prop_checkBatsTestDoesNotUseNegation2 = verify checkBatsTestDoesNotUseNegation "
 prop_checkBatsTestDoesNotUseNegation3 = verify checkBatsTestDoesNotUseNegation "#!/usr/bin/env bats\n@test \"name\"{ run ! true }"
 checkBatsTestDoesNotUseNegation params t =
     case t of
-        T_BatsTest _ _ (T_BraceGroup _ commands) -> mapM_ check (dropLast commands)
-            T_Banged id (T_Pipeline _ _ [T_Redirecting _ _ (T_Condition idCondition _ _)]) -> 
-                                err id 2293  
-                                "bats: ! <command> will never fail the test. Fold the `!` into the conditional!"
-            T_Banged id cmd -> errWithFix id 2293 
-                                "bats: ! <command> will never fail the test."
-                                (fixWith [replaceStart id params 0 "run "])
-            _ -> return ()
-    dropLast t =
+        T_BatsTest _ _ (T_BraceGroup _ commands) -> mapM_ (check commands) commands
+        _ -> return ()
+  where
+    check commands t =
         case t of
-            [_] -> []
-            x:rest -> x : dropLast rest
-            _ -> []
+            T_Banged id (T_Pipeline _ _ [T_Redirecting _ _ (T_Condition idCondition _ _)]) -> 
+                                if t `isLastOf` commands
+                                then style id 2315 "In bats, ! will not fail the test if it is not the last command anymore. Fold the `!` into the conditional!"
+                                else err id 2315
+                                        "In bats, ! does not cause a test failure. Fold the `!` into the conditional!"
+                                    
+            T_Banged id cmd -> if t `isLastOf` commands
+                                then styleWithFix id 2314
+                                                "In bats, ! will not fail the test if it is not the last command anymore. Use `run ! ` (on bats >= 1.5.0) instead."
+                                                (fixWith [replaceStart id params 0 "run "])
+                                else errWithFix id 2314
+                                                "In bats, ! does not cause a test failure. Use 'run ! ' (on bats >= 1.5.0) instead."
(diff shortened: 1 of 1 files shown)


koalaman:
Looks pretty reasonable! I only have minor comments:
* The message "bats: ! <command> will never fail the test." could be a more actionable "In bats, ! <command> does not cause a test failure. Use 'run ! <command>' instead."
* While the errors are similar, you might as well use two different error codes. That way, when people look at the wiki they'll immediately get the relevant one of the two suggested solutions (ps: sorry about overwriting the wiki page! The contents is still available in the edit history)
* I can shove the dropLast function into ASTLib or something so that the two functions that use it can share it

Do you have any idea about the false positive rate, such as whether `! cmd; [ $? -eq 0 ]` is a common idiom?

I also see that the Ubuntu version is a relatively old 1.2.1 where `run ! true` does not cause a failure. Do you have a sense of how long it takes before the new version is widely adopted? (It's probably less than ShellCheck's release cycle)

martin-schulze-vireso:
> The message "bats: ! <command> will never fail the test." could be a more actionable "In bats, ! <command> does not cause a test failure. Use 'run ! <command>' instead."

Okay, I thought I should not duplicate the edit suggestion.

> While the errors are similar, you might as well use two different error codes. That way, when people look at the wiki they'll immediately get the relevant one of the two suggested solutions (ps: sorry about overwriting the wiki page! The contents is still available in the edit history)

Okay, I must admit I tried to be lazy here...

> I can shove the dropLast function into ASTLib or something so that the two functions that use it can share it

I did not know where to put it, so I just let it sit there. I am fine with that.

> Do you have any idea about the false positive rate, such as whether ! cmd; [ $? -eq 0 ] is a common idiom?

I haven't done a survey on the frequency of this mistake, yet. I believe that most people are unaware of the problem since knowing it requires arcane bash knowledge paired with bats internal implementation knowledge. I have seen several instances in the bats self test suite and none of them explicitly checked the exit code.

> I also see that the Ubuntu version is a relatively old 1.2.1 where run ! true does not cause a failure. Do you have a sense of how long it takes before the new version is widely adopted? (It's probably less than ShellCheck's release cycle)

It is very hard to estimate. We have a following routes of "installation": via `git submodule`, docker image, npm, brew, distribution repository. Historically, bats had very sporadic releases which led some people to just pull master in as `git submodule`. We don't have telemetry that could tell us how many people are using which version so we really can't tell. All I can say is: there are people who upgrade to the current release on the day we release it and there are people who will run into errors on master on the day we push them.

koalaman:
`quicktest` keeps breaking with every new major release of cabal, presumably because I'm not sure how it all works. See https://github.com/koalaman/shellcheck/issues/2286 for the workarounds I last verified.

martin-schulze-vireso:
Okay, I did a little survey of all '*.bats' files in all projects that were listed as dependent on bats. These were ~1.5k files with ~90 KLOC. I found only two beginner projects that made this mistake. I added some more, widely used projects like https://github.com/kata-containers/tests
https://github.com/dokku/dokku  https://github.com/dolthub/dolt most of the established projects use bats-assert's `refute` or `! <command> || false`.

koalaman:
Sounds good! I'm happy to merge this at any time.

martin-schulze-vireso:
Okay, I split the error codes into two and updated them to use [SC2314](../wiki/SC2314)/[SC2315](../wiki/SC2315).

koalaman:
Oops, merged too soon and force pushed off. The tests don't pass anymore. Can you have a look?

martin-schulze-vireso:
Where did the tests fail? Do you have a link?

koalaman:
It failed for me when testing locally. Does the latest iteration pass for you?

martin-schulze-vireso:
No, I must admit that I did not run the tests because I had trouble setting up the haskell environment again and decided to rely on  CI, which does not run on PRs it seems?

Anyways, it took me another hour to set up some environment that runs the tests and can confirm they fail. But it is beyond my why. They look correct and my Haskell skill is too weak to debug this.

martin-schulze-vireso:
I worked around that by creating a file with the test code, thereby getting better diagnostics.

PR #2289, opened 2021-07-31
Minor changes in README

nafigator:
It's a recommended practice to use apt instead `apt-get`:

> apt is a second command-line based front end provided by APT which overcomes some design mistakes of apt-get.

https://debian-handbook.info/browse/stable/sect.apt-get.html

Also added `sudo` for commands needed root privileges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ On systems with Stack (installs to `~/.local/bin`):
 
 On Debian based distros:
 
-    apt-get install shellcheck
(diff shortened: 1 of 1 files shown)


koalaman:
Looks good. Is it still the case that FreeBSD does not come with `sudo` by default? If so, maybe that one should be left alone

nafigator:
@koalaman You're right. FreeBSD section fixes removed.

Issue #2054, opened 2020-09-08
Enhancement: `set -e` is ignored in functions when used in the "if" context


mvo5:
First let me say THANK YOU for the wonderful shellcheck tool. I use it daily and love it.

#### For new checks and feature suggestions
- [x] https://www.shellcheck.net/ (i.e. the latest commit) currently gives no useful warnings about this
- [x] searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

I found https://github.com/koalaman/shellcheck/issues/1484 which is slightly similar but also different enough I feel to justify it's own issue. If I'm wrong please just close as duplicate.

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/sh
set -e

foo() {
    set -e
    false
    echo "output after failed command even with set -e"
}

if foo; then
	echo
fi
```

#### Here's what shellcheck currently says:

$ shellcheck myscript
No issues detected!

#### Here's what I wanted or expected to see:

It would be nice if it would warn that if shell functions are used inside an "if" any `set -e` is disabled while the function runs. I added the "set -e" in foo() just for demonstration purposes, it's does not change anything.

marc-h38:
If the exit code of `foo` is sometimes tested with `if foo` and elsewhere not, would you warn or not?

kerolasa:
I played with the issue a bit, and got myself more confused. `set -e` is not inherited it needs to be set again within function. Okay, But `set -E` does not seem to be inherited either. If I add a trap within function (uncomment below line to see effect) it will be active later outside function scope - where return is not right way to exit.

```
#!/bin/bash
set -e # exit on errors
set -E # inherit ERR trap
trap 'echo trap; exit 1' ERR

foo() {
        set -e
        #trap 'echo function trap; return 1' ERR
        false
        echo "output after failed command even with set -e"
        return 0
}

if foo; then
        echo "still going"
        false
fi
```
Tested using bash version 5.0.18(1)-release.

kurahaupo:
Functions can be nested in un-obvious ways; consider:

```sh
set -e

f() {
    do_something  # is this exempt?
}
g() {
    echo starting "$1"
    "$@"
}

if g f
then echo Happy
else echo Sad
fi
```
or
```sh
set -e

f() {
    do_something  # is this exempt?
}
g() {
    echo starting "$1"
    if "$@"
    then r=$? # 0
    else r=$?
    fi
    echo finished "$1"
    return "$r"
}

g f
```

[...]

Issue #1484, opened 2019-02-03
Enhancement: `set -e` is ignored in a subshell with specific surrounding context


cuihaoleo:
#### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/bin/bash
(
    set -e
    false
    echo AAA
) && echo BBB
```

#### Here's what shellcheck currently says:

Nothing.

#### Here's what I wanted or expected to see:

See: https://unix.stackexchange.com/questions/65532/why-does-set-e-not-work-inside-subshells-with-parenthesis-followed-by-an-or

> The -e setting shall be ignored when executing the compound list following the while, until, if, or elif reserved word, a pipeline beginning with the ! reserved word, or any command of an AND-OR list other than the last.

It is a really counter-intuitive behavior. Shellcheck may warn user about ignored `set -e`.

PR #2303, opened 2021-08-22
Show info about `set -e` suppression during function calls

DoxasticFox:
This PR provides an enhancement very similar to the one described in #2054. The following example is copied from that issue:

```bash
#!/bin/bash
set -e

foo() {
    set -e
    false
    echo "output after failed command even with set -e"
}

if foo; then
    echo
fi
```

The current version of `shellcheck` provides no notice, although this PR produces the following:

```
if foo; then
   ^-^ SC2304: This function invocation will run with set -e disabled.

For more information:
  https://www.shellcheck.net/wiki/SC2304 -- This function invocation will run...
```

It also handles all the other cases (that I know of) where `set -e` can be disabled during a function call. The new notice can be suppressed by inserting `set -e` into the correct context. For example this program produces an error in `shellcheck`:

```bash
#!/bin/bash
set -e
foo() { :; }
bar=$(foo)
```

But this one doesn't:

```bash
#!/bin/bash
set -e
foo() { :; }
bar=$(set -e; foo)
```

Note that Bash doesn't allow `set -e` to be "re-enabled" in most contexts. For example, this wouldn't re-enable `set -e`:

```bash
#!/bin/bash
set -e
foo() { :; }
if
    set -e
    foo
then
    :
fi
```

As a result, `shellcheck` would still display the error. Another interesting edge case is this:

```bash
#!/bin/bash
set -e
foo() { :; }
bar=$(set -e; foo) || true
```

Although you can ordinarily re-enable `set -e` in command substitutions, the `|| true` overrides that. So a `shellcheck` notice is still produced in this case.

It's also worth highlighting that this PR only deals provides notices about _function_ invocations, so `shellcheck` wouldn't produce any output for the program `set -e; x=$(git)`. Related issues and PRs exist (#1484, #2237) to address more general cases.

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -253,6 +253,13 @@ optionalTreeChecks = [
         cdPositive = "[ -e /etc/issue ]",
         cdNegative = "[[ -e /etc/issue ]]"
     }, checkRequireDoubleBracket)
+
+    ,(newCheckDescription {
+        cdName = "check-set-e-suppressed",
+        cdDescription = "Notify when set -e is suppressed during function invocation",
+        cdPositive = "set -e; func() { cp *.txt ~/backup; rm *.txt; }; func && echo ok",
+        cdNegative = "set -e; func() { cp *.txt ~/backup; rm *.txt; }; func; echo ok"
+    }, checkSetESuppressed)
     ]
 
 optionalCheckMap :: Map.Map String (Parameters -> Token -> [TokenComment])
@@ -392,6 +399,24 @@ replaceToken id params r =
 
 surroundWith id params s = fixWith [replaceStart id params 0 s, replaceEnd id params 0 s]
 fixWith fixes = newFix { fixReplacements = fixes }
+
+analyse f t = execState (doAnalysis f t) []
+
+-- Make a map from functions to definition IDs
+functions t = Map.fromList $ analyse findFunctions t
+findFunctions (T_Function id _ _ name _)
+    = modify ((name, id):)
+findFunctions _ = return ()
+
+-- Make a map from aliases to definition IDs
+aliases t = Map.fromList $ analyse findAliases t
+findAliases t@(T_SimpleCommand _ _ (_:args))
+    | t `isUnqualifiedCommand` "alias" = mapM_ getAlias args
+findAliases _ = return ()
+getAlias arg =
+    let string = onlyLiteralString arg
+    in when ('=' `elem` string) $
+        modify ((takeWhile (/= '=') string, getId arg):)
 
 prop_checkEchoWc3 = verify checkEchoWc "n=$(echo $foo | wc -c)"
 checkEchoWc _ (T_Pipeline id _ [a, b]) =
@@ -2239,21 +2264,11 @@ checkFunctionsUsedExternally params t =
         findExecFlags = ["-exec", "-execdir", "-ok"]
         dropFlags = dropWhile (\x -> "-" `isPrefixOf` fst x)
 
-    -- Make a map from functions/aliases to definition IDs
-    analyse f t = execState (doAnalysis f t) []
-    functions = Map.fromList $ analyse findFunctions t
-    findFunctions (T_Function id _ _ name _) = modify ((name, id):)
-    findFunctions t@(T_SimpleCommand id _ (_:args))
-        | t `isUnqualifiedCommand` "alias" = mapM_ getAlias args
-    findFunctions _ = return ()
-    getAlias arg =
-        let string = onlyLiteralString arg
-        in when ('=' `elem` string) $
-            modify ((takeWhile (/= '=') string, getId arg):)
+    functionsAndAliases = Map.union (functions t) (aliases t)
 
     checkArg cmd (_, arg) = sequence_ $ do
         literalArg <- getUnquotedLiteral arg  -- only consider unquoted literals
-        definitionId <- Map.lookup literalArg functions
+        definitionId <- Map.lookup literalArg functionsAndAliases
         return $ do
             warn (getId arg) 2033
               "Shell functions can't be passed to external commands."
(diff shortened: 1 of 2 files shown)


src/ShellCheck/Analytics.hs
```diff
@@ -393,6 +431,18 @@ replaceToken id params r =
...
+functions t = Map.fromList $ analyse findFunctions t
+findFunctions (T_Function id _ _ name _) = modify ((name, id):)
+findFunctions t@(T_SimpleCommand id _ (_:args))
+    | t `isUnqualifiedCommand` "alias" = mapM_ getAlias args
+findFunctions _ = return ()
+getAlias arg =
+    let string = onlyLiteralString arg
+    in when ('=' `elem` string) $
+        modify ((takeWhile (/= '=') string, getId arg):)
+
```
DoxasticFox, line 445: This snippet was moved from here:

https://github.com/koalaman/shellcheck/blob/da7b28213ef65ef4a0d702831d1d358f37e4f114/src/ShellCheck/Analytics.hs#L2244-L2252

The only difference is that `functions` was made to take an argument `t`.

src/ShellCheck/Analytics.hs
```diff
@@ -4583,6 +4623,45 @@ checkArrayValueUsedAsIndex params _ =
...
 
+prop_checkDisabledSetEInIf1  = verifyTree    checkDisabledSetEInIf "set -e; f(){:;}; x=$(f)"
+prop_checkDisabledSetEInIf2  = verifyNotTree checkDisabledSetEInIf "f(){:;}; x=$(f)"
+prop_checkDisabledSetEInIf3  = verifyNotTree checkDisabledSetEInIf "set -e; f(){:;}; x=$(set -e; f)"
+prop_checkDisabledSetEInIf4  = verifyTree    checkDisabledSetEInIf "set -e; f(){:;}; f || echo"
+prop_checkDisabledSetEInIf5  = verifyNotTree checkDisabledSetEInIf "set -e; f(){:;}; echo || f"
+prop_checkDisabledSetEInIf6  = verifyTree    checkDisabledSetEInIf "set -e; f(){:;}; baz=$(set -e; f) || :"
+prop_checkDisabledSetEInIf7  = verifyNotTree checkDisabledSetEInIf "set -e; f(){:;}; baz=$(echo \"\") || :"
+prop_checkDisabledSetEInIf8  = verifyTree    checkDisabledSetEInIf "set -e; f(){:;}; f && echo"
+prop_checkDisabledSetEInIf9  = verifyTree    checkDisabledSetEInIf "set -e; f(){:;}; f || echo"
```
uri-canva, line 4634: This is the same as `prop_checkDisabledSetEInIf4`.

src/ShellCheck/Analytics.hs
```diff
@@ -4583,6 +4623,45 @@ checkArrayValueUsedAsIndex params _ =
...
+prop_checkDisabledSetEInIf10 = verifyNotTree checkDisabledSetEInIf "set -e; f(){:;}; echo && f"
+prop_checkDisabledSetEInIf11 = verifyNotTree checkDisabledSetEInIf "set -e; f(){:;}; echo || f"
```
uri-canva, line 4636: This is the same as `prop_checkDisabledSetEInIf5`.

src/ShellCheck/Analytics.hs
```diff
@@ -4583,6 +4623,45 @@ checkArrayValueUsedAsIndex params _ =
...
+                    mapM_ (checkArg name) argv
+                _ -> return ()
+    checkCommand _ _ = return ()
+
+    functions_ = functions t
+
+    checkArg cmd arg = sequence_ $ do
+        literalArg <- getUnquotedLiteral arg
+        definitionId <- Map.lookup literalArg functions_
+        return $ info (getId arg) 2304 "This function invocation will run with set -e disabled."
```
uri-canva, line 4663: You're not using `functions_`, `literalArg` or `definitionId` for anything here. The check you copied this code from had a message that included both the function definition and the function call site, that's why it used the function definition map, but this check's message only references the function call site.

Having both might be helpful if aliases are used, so that even if the call site uses an alias, the function definition can be highlighted to show why it was detected as a function alias and not a command alias.

uri-canva:
Given that it's very idiomatic to have functions that are meant to be called in conditionals even in scripts where `set -e` is used, it's best to make this an optional check by adding it to `optionalTreeChecks`.

src/ShellCheck/Analytics.hs
uri-canva: Ignore this, I forgot you need to call `Map.lookup` just to be able to tell if `arg` is a function call or a command call.

DoxasticFox: I need to rewrite `checkArg` not to take an unused `cmd` parameter, so there's certainly still room for improvement here.

DoxasticFox:
I ran this on the monorepo at work and generated 318 new shellcheck notices! Most of them look useful to me, but there's some useless ones too.

I'd say the most common useless class of error is for library functions which are clearly safe. It'd be nice if we could add a shellcheck directive to the _function_ to mark it as being safe for use in `set -e`-suppressed contexts. Unfortunately, although I haven't investigated yet, I assume you either need to add the directive near where the error occurs, or make significant changes to shellcheck.

There's also a few instances where function invocations end in `|| true`. My initial thought was that it was useless because the author clearly knew what they were doing in that case. But on second thought, maybe they assumed `|| true` only suppressed the function's return code, as opposed to potentially changing the code path it follows.

There's also a few cases I missed. For example, it doesn't notify about subshells invoked with parentheses. Also, it _does_ notify about functions which are passed as arguments. I got a few false positives that way.

uri-canva:
> For example, it doesn't notify about subshells invoked with parentheses.

Those always inherit `set -e` no?

```
bash-4.4$ (set -e; (a() { echo hello; false; echo world; } ; a))
hello
bash-4.4$ (set -e; echo $(a() { echo hello; false; echo world; } ; a))
hello world
```

DoxasticFox:
@koalaman I think I'm done making changes. PTAL when you're ready.

There's a few things I'd like to talk about, centred around the phrasing of the new check (SC2304). I've noticed that, unlike this new check, the vast majority of existing ones don't say what's wrong with a code snippet. Rather, they say how it can be improved. I didn't do that, because I think there's a lot of ambiguity around the best course of action. [All the cases](https://github.com/koalaman/shellcheck/blob/068b8894f58aaccb74e28ca99fef97399170fc41/src/ShellCheck/Analytics.hs#L361-L372) covered by SC2304 can be fixed by turning the affected functions into scripts, but I imagine that many people would consider that excessive. We can provide more nuanced feedback for the cases where the Bash programmer can simply re-enable `set -e`, i.e. in dollar or backtick command substitutions. In those cases, we can provide an error message like `Add 'set -e' here or 'shopt -s inherit_errexit' to the top of your script to ensure this function runs with 'set -e' enabled`. But in all the other cases, I think the programmer will have to decide what's best for their situation. And actually, I'm not totally sold on telling the user what to do for command substitutions, because `hasSetE` is `True` if `set -e` is present _anywhere_ in the script, which risks false positives. AFAIK, there's no way to know if `set -e` was set in a `Token`'s parent context until the Bash script is executed. So I think it might be better for `shellcheck` to point out the potential mistake rather than advocate a potentially misguided solution. Strategies for dealing with SC2304 can be described in the wiki. What do you think?

src/ShellCheck/Analytics.hs
```diff
@@ -4583,6 +4593,101 @@ checkArrayValueUsedAsIndex params _ =
...
+prop_checkSetESuppressed16 = verifyTree    checkSetESuppressed "set -e; f(){ :; }; until set -e; f; do :; done"
+prop_checkSetESuppressed17 = verifyTree    checkSetESuppressed "set -e; f(){ :; }; cat <(f)"
+prop_checkSetESuppressed18 = verifyTree    checkSetESuppressed "set -e; f(){ :; }; cat <(set -e; f)"
+prop_checkSetESuppressed19 = verifyNotTree checkSetESuppressed "set -e; f(){ :; }; g(){ :; }; g f"
+prop_checkSetESuppressed20 = verifyNotTree checkSetESuppressed "set -e; shopt -s inherit_errexit; f(){ :; }; x=$(f)"
+checkSetESuppressed params t =
+    if hasSetE params then runNodeAnalysis checkNode params t else []
+  where
+    checkNode _ (T_SimpleCommand _ _ (cmd:_)) = when (isFunction cmd) (checkCmd cmd)
+    checkNode _ _ = return ()
```
koalaman, line 4620: Do you think this warning should trigger for things like `{ a; b; } || c` and `$(a; b)` for the same reason why it triggers for `f() { a; b; }; f || c` and `$(f)`?

It's just a thought and I don't expect it of this PR.

src/ShellCheck/Analytics.hs
```diff
@@ -4583,6 +4593,101 @@ checkArrayValueUsedAsIndex params _ =
...
+                T_UntilExpression _ conditions _ ->
+                    Just (conditions, sc2304 "an 'until' condition")
+                T_WhileExpression _ conditions _ ->
+                    Just (conditions, sc2304 "a 'while' condition")
+
+                -- Process substitution is weird. The body of the function
+                -- inherits `set -e`, but if that function fails, the script
+                -- doesn't exit.
+                T_ProcSub _ _ cmds ->
+                    Just (cmds, sc2305)
```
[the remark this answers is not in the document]
DoxasticFox: That's a good point. It's not `set -e` being suppressed, but the exit code.

I'd still like to raise a separate PR for the process substitution case, later on. I should give it an SCxxxx number separate from the cases you mention, right? After all, [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) already exists for one related case.

src/ShellCheck/Analytics.hs
```diff
@@ -393,6 +401,18 @@ replaceToken id params r =
...
+analyse f t = execState (doAnalysis f t) []
+functions t = Map.fromList $ analyse findFunctions t
+findFunctions (T_Function id _ _ name _) = modify ((name, id):)
+findFunctions t@(T_SimpleCommand _ _ (_:args))
+    | t `isUnqualifiedCommand` "alias" = mapM_ getAlias args
+findFunctions _ = return ()
+getAlias arg =
+    let string = onlyLiteralString arg
+    in when ('=' `elem` string) $
+        modify ((takeWhile (/= '=') string, getId arg):)
```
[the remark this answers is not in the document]
DoxasticFox: Ignoring aliases SGTM, but I didn't do the optional bit because I worry it'll be brittle.

src/ShellCheck/Analytics.hs
```diff
@@ -4583,6 +4593,101 @@ checkArrayValueUsedAsIndex params _ =
...
+    isFunction cmd = isJust $ do
+        literalArg <- getUnquotedLiteral cmd
+        Map.lookup literalArg functions_
+
+    suppressionReasons t =
+        map snd $ catMaybes $ go (getPath (parentMap params) t)
+        where
+            go [] = []
+            go [_] = []
+            go (child:parent:rest) =
```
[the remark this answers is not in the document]
DoxasticFox: Thanks!

src/ShellCheck/Analytics.hs
DoxasticFox: When I ran this PR on the monorepo at work, I found that `set -e` was mostly suppressed in functions which were used in command substitutions and `if` statements, not `||` or `&&` conditions. (I think that's because people don't want to branch with `&&` and `||` in production code.) So, admittedly, I don't have much empirical evidence to suggest it'll be _that_ useful.

I also think `set -e`'s behaviour is more insidious for functions than the cases you mention, because, one might reasonably expect functions and commands to respond similarly to `set -e`. i.e. You wouldn't expect the behaviour of the command/function to change. You'd expect the script's response to the command/function's exit code to change.

Nonetheless, I still think the cases you mention are unintuitive enough to warrant their own warnings, because nobody could know `errexit` is disabled in those contexts without being told. I don't really want to raise a PR though because I haven't seen those pop up much in the wild and I'm lazy :joy:

koalaman:
Looks good to me! Please squash into one neat commit for merge.

The checks are helpful for all shells, but the optional description is bash specific. I'll probably switch it out for something broader that also highlights the problem, maybe:

```
    ,(newCheckDescription {
        cdName = "check-set-e-suppressed",
        cdDescription = "Notify when set -e is suppressed during function invocation",
        cdPositive = "set -e; func() { cp *.txt ~/backup; rm *.txt; }; func && echo ok",
        cdNegative = "set -e; func() { cp *.txt ~/backup; rm *.txt; }; func; echo ok"
    }, checkSetESuppressed)
```

I might potentially also add support for [using `# shellcheck disable` on the function](https://github.com/koalaman/shellcheck/blob/747bd8fd6af31f578bf8683ac9df9f29d5ca6d0b/src/ShellCheck/Analytics.hs#L2840) instead of having to do it on every call site when you know the function is safe.

DoxasticFox:
Thanks, @koalaman!

In addition to squashing the current commits, I made two more changes:

* Added the description you mentioned.
* Bumped the SCxxxx numbers because they collide with some existing checks now. Might be worthwhile triple-checking them before you merge.

>I might potentially also add support for using # shellcheck disable on the function

You read my mind! Our codebase at work has some functions which are obviously safe to use with `errexit` suppressed. I mentioned to @uri-canva while writing this PR that suppressing errors like this would be nice, but I didn't implement it because I didn't realise it was that easy. Still, I'm enough of a Haskell n00b that I'd expect a few too many iterations on the PR before getting it right :joy:. Better leave it to the experts for now :smile:

It'd still like to raise a PR for https://github.com/koalaman/shellcheck/pull/2303#discussion_r698056295 though, because it seems pretty simple and I think it's a common mistake.

uri-canva:
Nice! Thanks @DoxasticFox and @koalaman for the quick review!

PR #2318, opened 2021-09-02
Do not suggest `grep -c` as a replacement for `grep -l/-L | wc -l`

FabianWolff:
`grep -c` counts the number of matches, and `grep -l/-L` outputs the names of the files with/without matches. Now, if I want the number of files with/without matches, I would write something like this:
```sh
grep -li "fail" ./*.log | wc -l
```
However, Shellcheck suggests replacing the `grep -l/-L | wc -l` with `grep -c`, which is not the correct solution:
```
Line 3:
grep -li "fail" ./*.log | wc -l
^-- SC2126: Consider using grep -c instead of grep|wc -l.
```
Therefore, I have added `-l/-L` to the blacklist of `grep` options that prevent [SC2126](https://github.com/koalaman/shellcheck/wiki/SC2126) from being issued.


koalaman:
Thanks!

Issue #1733, opened 2019-10-26
False negative in SC2155


lvjp:
#### For bugs
- Rule Id: SC2155
- My shellcheck version: 0.7.0
- [x] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2155)
- [x] I tried on shellcheck.net and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh
#!/usr/bin/env bash
set -o errexit
export FOO="$(myEnvBuilder)"
FOO="$(myEnvBuilder)" ./myCmd.sh
```
#### Here's what shellcheck currently says:

```
In myscript.sh line 3:
export FOO="$(myEnvBuilder)"
       ^-^ SC2155: Declare and assign separately to avoid masking return values.
```

#### Here's what I wanted or expected to see:

```
In myscript.sh line 3:
export FOO="$(myEnvBuilder)"
       ^-^ SC2155: Declare and assign separately to avoid masking return values.

In myscript.sh line 4:
FOO="$(myEnvBuilder)" ./myCmd.sh
^-^ SC2155: Declare and assign separately to avoid masking return values.
```

In the line 4, the return value of `myEnvBuilder` is effectively ignored but not detect as a bug within shellcheck.

matthewpersico:
Can you simplify the sample? I am betting that the ‘export’ has nothing to do with it; it’s probably the addition of the string that confuses shellcheck. What happens if you eliminate the export line totally (I’d try it myself but I’m on my iPhone in an IHOP).

lvjp:
@matthewpersico I edit the issue to have a simpler example.

lvjp:
Thank you @jabberabbe, you commit jabberabbe/shellcheck@be0d0247b180f0da68edac57ae1498b90eb2ff30 did the job.

Can't wait for pull request #1739 to be approuved.

koalaman:
What's your suggested rewrite for this?

PR #2320, opened 2021-09-05
Add extra checks for masked return codes

DoxasticFox:
Adds the check mentioned here, plus related ones: https://github.com/koalaman/shellcheck/pull/2303#discussion_r698056295 Example:

`/tmp/a.sh`:

```bash
#!/usr/bin/env bash
set -e
readarray -t files < <(ls)
echo "${files[@]}"
```

```
% ./quickrun /tmp/a.sh --enable=check-extra-masked-returns
Up to date

In /tmp/a.sh line 3:
readarray -t files < <(ls)
                       ^-- SC2312: Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore).

For more information:
  https://www.shellcheck.net/wiki/SC2312 -- Consider invoking this command se...

```

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -260,6 +260,13 @@ optionalTreeChecks = [
         cdPositive = "set -e; func() { cp *.txt ~/backup; rm *.txt; }; func && echo ok",
         cdNegative = "set -e; func() { cp *.txt ~/backup; rm *.txt; }; func; echo ok"
     }, checkSetESuppressed)
+
+    ,(newCheckDescription {
+        cdName = "check-extra-masked-returns",
+        cdDescription = "Check for additional cases where exit codes are masked",
+        cdPositive = "rm -r \"$(get_chroot_dir)/home\"",
+        cdNegative = "set -e; dir=\"$(get_chroot_dir)\"; rm -r \"$dir/home\""
+    }, checkExtraMaskedReturns)
     ]
 
 optionalCheckMap :: Map.Map String (Parameters -> Token -> [TokenComment])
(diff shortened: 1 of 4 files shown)


DoxasticFox:
CC @uri-canva

koalaman:
I think this would need to cover quite a bit more to live up to its name, such as `false | true`, `true $(false)`, and `x=$(false)$(true)`, and a check specifically for process substitutions in scripts containing `set -e` seems a bit narrow on its own.

DoxasticFox:
@koalaman I think it highlights another issue with my chosen name: It's that the check isn't related to `set -e` so much as it is to masking return values (e.g. as in [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155)). So although `set -e` makes Bash's treatment of the masked exit codes even more unexpected, I'd still say it's not the core issue. How about I rename `check-set-e-ignored` to `check-masked-exit-codes`, or `check-extra-masked-exit-codes` (seeing as `shellcheck` already deals with one case of masking in SC2155). I'll also add checks for the cases you mention, being careful to avoid the cases already handled by SC2155. SGTY?

koalaman:
SGTM. In this case it should recognize `|| true` so that you have an easy escape hatch for things you're willing to assume will succeed, such as `x $(y || true)` and `x < <(y || true)`.

DoxasticFox:
@koalaman PTAL

src/ShellCheck/Analytics.hs
```diff
@@ -4734,5 +4742,90 @@ checkSetESuppressed params t =
...
+prop_checkExtraMaskedReturns3  = verifyTree    checkExtraMaskedReturns "false | true"
+prop_checkExtraMaskedReturns4  = verifyNotTree checkExtraMaskedReturns "set -o pipefail; false | true"
+prop_checkExtraMaskedReturns5  = verifyNotTree checkExtraMaskedReturns "false | true || true"
+prop_checkExtraMaskedReturns6  = verifyTree    checkExtraMaskedReturns "true $(false)"
+prop_checkExtraMaskedReturns7  = verifyTree    checkExtraMaskedReturns "x=$(false)$(true)"
+prop_checkExtraMaskedReturns8  = verifyTree    checkExtraMaskedReturns "true <<<$(false)"
+prop_checkExtraMaskedReturns9  = verifyNotTree checkExtraMaskedReturns "echo asdf | false"
+prop_checkExtraMaskedReturns10 = verifyNotTree checkExtraMaskedReturns "readonly x=$(false)"
+prop_checkExtraMaskedReturns11 = verifyTree    checkExtraMaskedReturns "readarray -t files < <(ls)"
+checkExtraMaskedReturns params t = runNodeAnalysis checkNode params t
```
koalaman, line 4756: So far so good! Here are some cases it misses:

False negative:
```
x="$(false)$(false)"      # Check out getWordParts :: Token -> [Token]
x="$(false)""$(false)"    # It takes a T_NormalWord and flattens it into a list of all the quoted and unquoted elements in the string
x=( $(false) $(false) )
x=( $(false) [4]=$(false) )

# These may be less relevant:
cat << foo
 $(false)
foo
x=$(( $(date +%s) ))
[[ $(false) ]]
```

False positive:
```
x=$(false) y=z         # T_SimpleCommand has two assignments in its list in this case
x=$(false) y=$(false)  # Only the first one is masked
```

These are already correctly handled but untested:
```
echo $(( $(date +%s) ))
x=`false``false`
x=( $(false) )
```

DoxasticFox: Thanks for all these test cases. Any suggestion as to how I should approach this?

```bash
x=$(false) y=$(false)  # Only the first one is masked
```

I was thinking that once I find a `T_SimpleCommand`, I'd traverse the tree upwards, looking for a `T_Assignment`. Once I find that, I'd look at the assignments on the the left of the `T_SimpleCommand` and check to see if their descendants contain any `T_SimpleCommand` nodes. The trouble is that it's not obvious to me how I can iterate over the descendants of a node without having a huge `case ... of` statement which essentially handles each rule in Bash's grammar.

koalaman: This big case...of statement already exists. There's a `doAnalysis` which invokes each node in a tree with a monadic action. The obvious monad is Cont, but to avoid importing that, perhaps Maybe:

```
containsSimpleCommand rootnode = isNothing $ doAnalysis (\c -> case c of T_SimpleCommand {} -> fail ""; _ -> return ()) rootnode
```

Currently the code finds the potentially masked T_SimpleCommand, and tries to determine if it is masked. This has some issues like `echo $(t=$(false); echo $t)` emitting two warnings even though only one is actually masked, and `echo $([[ x ]])` emits nothing since it's a compound command instead of a simple command.

An alternative approach is finding the potential masker, e.g. a T_ProcSub, or a `T_SimpleCommand _ [] (cmd:args)` where `any isCommandSubstitution $ concatMap getWordParts args`. Since it seems like this is basically required to recognize the context anyways, maybe this is easier.

DoxasticFox:
@koalaman How am I going?

DoxasticFox:
Looks like this PR will address https://github.com/koalaman/shellcheck/issues/1733. Not sure if that's a good thing though; is it okay for it to be an optional check?

koalaman (commented):
Looks pretty good and feature complete to me!

Did you run this on a larger codebase? What was your experience?

DoxasticFox:
>Did you run this on a larger codebase? What was your experience?

@koalaman Yeah, that's why I added these tests:

https://github.com/koalaman/shellcheck/blob/56ddb5feb5b35e46784a0b7a647c7ce02fcc97e2/src/ShellCheck/Analytics.hs#L4772-L4775

But actually I should run it once more and eyeball the results again now that I've made those tweaks.

Also, to make the test `SCRIPT_DIR=\"$( cd \"$( dirname \"$0\" )\" && pwd )\"` pass, I decided to make `dirname` a "harmless" command. What I really would've liked was to make it harmless only when invoked on `$0` or `${BASH_SOURCE[0]}`, but I couldn't see an easy and obvious way to do that. Got any pithy ways to check for that, oh Haskell master?

koalaman:
> What I really would've liked was to make it harmless only when invoked on $0 or ${BASH_SOURCE[0]}

I would avoid this kind of overfitting TBH. It's often better for the test to be predictable, even if it makes it slightly more inaccurate. "dirname and basename are ignored since they're rarely sources of failure" is simple and easy, while it's harder to build a mental model given observations like:

* No warning: `basename ${BASH_SOURCE[0]}`
* Warning: `basename ${BASH_SOURCE[0]?}`
* No warning: `basename $0`
* Warning: `current_exe=$0; basename $current_exe`
* No warning: `dirname ${BASH_SOURCE[0]}`
* Warning: `dirname ${BASH_SOURCE[-1]}`

Other than that it looks good to me. Squash for merge?

PS: Any thoughts on a more convincing doc example? Maybe something like

        cdPositive = "rm -r \"$(get_chroot_dir)/home\"",
        cdNegative = "set -e; dir=\"$(get_chroot_dir)\"; rm -r \"$dir/home\""

DoxasticFox:
Squashed

>It's often better for the test to be predictable, even if it makes it slightly more inaccurate.

Makes sense. I reverted 77f393e (the commit to specifically silence `dirname $0` and `basename $0`) and added `basename` to the list of harmless commands instead.

>PS: Any thoughts on a more convincing doc example? Maybe something like [...]

:+1:

koalaman:
Thanks! I've merged it and added a wiki page and changelog entry.

DoxasticFox:
You might also want to consider closing https://github.com/koalaman/shellcheck/issues/1733

koalaman:
@DoxasticFox Sure. There are a few minor issues I'd like to squeeze in first, including #2362 to extend this warning to `local -r x=$(false)` which you might be interested in, but I can probably get a new version out next weekend if that works.

koalaman:
@DoxasticFox v0.8.0 is now out

PR #2352, opened 2021-10-07
Allow specifying color= in shellcheckrc (closes #2350)

### LINKED ISSUE #2350: Also parse the `color` option from the config file
Shellcheck seems to ignore the `color=` option if it is placed in the .shellcheckrc file. IMO any commandline argument should be also parseable from the config file (the user can still override this after all).

This feature request is to support honoring the `color=` directive from the config file.

oliv3r:
To be able to set the color option in the config file (detection in GitLab CI fails for example) is useful when using template shellcheck containers for one. Secondly, storing config in the repo is also a thing :)

Sadly, I got a bit overconfident and got lost very quickly. I never seen haskell before so there's a learning curve required I suppose. I did make a start hopefully for whomever can pick it up :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Git (0.8.0)
 ### Added
+- `color=` directive can be added to .shellceckrc
 - `disable=all` now conveniently disables all warnings
 - `external-sources=true` directive can be added to .shellcheckrc to make
   shellcheck behave as if `-x` was specified.
--- a/shellcheck.1.md
+++ b/shellcheck.1.md
@@ -41,7 +41,10 @@ not warn at all, as `ksh` supports decimals in arithmetic contexts.
 
 :   For TTY output, enable colors *always*, *never* or *auto*. The default
     is *auto*. **--color** without an argument is equivalent to
-    **--color=always**.
+    **--color=auto**.
+
+    This option may also be enabled using `color=` in
+    `.shellcheckrc`. This flag takes precedence.
 
 **-i**\ *CODE1*[,*CODE2*...],\ **--include=***CODE1*[,*CODE2*...]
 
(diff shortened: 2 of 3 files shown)


Alhadis:
IMHO, this really belongs in an environment variable (`FORCE_COLOR=1` and [`NO_COLOR=1`](http://no-color.org/)). It's easier to forcefully enable/disable colour support for a process this way, and users typically do so cognisant of where/how output is being rendered; e.g., [an Emacs buffer](https://github.com/scottaj/mocha.el/pull/63) or pager like `less(1)` will want consistent colouration behaviour regardless of a project's local configuration.

ferdnyc:
I agree with both @koalaman and @Alhadis.

@koalaman makes an excellent point about the `.shellcheckrc` files. They're not appropriate for this type of per-invocation _preferences_ / _user default_ options, since they're locally scoped. If flags like `--color` and `--wiki-link-count` **were** going to be controllable via a configuration file, I think it should be a new, single-instance preferences file, with a default location like `$HOME/.config/shellcheck.conf` (though perhaps overridable via `shellcheck --config=<path>`). That file would be completely separate from the `.shellcheckrc` files, though.

**HOWEVER**, in terms of _`color`_, specifically, @Alhadis is right that supporting `FORCE_COLOR` and `NO_COLOR` environment variables would be a simpler and more widely supported /  understood solution. Respecting those envvars would likely make a file-based config for `--color` unnecessary.

...And the other side of the coin is, **_even if_** a file-based `color=` option is supported, it's **STILL** in the best interests of `shellcheck` to _also_ respect the (becoming-)standard environment variables for controlling color output.

The color-control envvars don't solve `wiki-link-count` & etc., of course. Though, an additional, generic envvar, something like `SHELLCHECK_ARGS`, could. e.g.:

```bash
# These three lines of shell input...
NO_COLOR=1
SHELLCHECK_ARGS="--format=checkstyle --wiki-link-count=2"
shellcheck -e SC1011 .

# Would have the exact same effect as this command...
shellcheck --color=never --format=checkstyle --wiki-link-count=2 \
  -e SC1011 .
```

PR #2388, opened 2021-11-26
Add --exit-zero option to exit with 0 in case of shell errors or warn…

khancyr:
…ings

Hello, 
I don't know anything on haskell. I tried my best but without understanding how it works ...
The idea is to have a new cli option to always return an exit 0 when we have warnings or errors. This allow to integrate shellchecker into CI and not beeing blocking (exit 1 means CI failure ...) and still output the errors.

The expected behavior is the same as the --exit-zero option on flake8 on Python.

Can somebody guide me to make this happen ?

### DIFF
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -74,7 +74,8 @@ data Options = Options {
     externalSources  :: Bool,
     sourcePaths      :: [FilePath],
     formatterOptions :: FormatterOptions,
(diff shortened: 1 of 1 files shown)


koalaman:
Thanks, but you can already do this with `shellcheck ... || true`. This works for all commands, not just ShellCheck.

PR #2414, opened 2021-12-14
Fix bug in 2126 when using after/before flags with grep

runejuhl:
Using `--after-context`/`-A` or `--before-context`/`-B` would give a warning
recommending the user to use `grep -c`, even though that would give a different
result than using `grep | wc -l`:

```fundamental
$ echo -e "1\n2\n3" | grep -cA 3 1
1
$ echo -e "1\n2\n3" | grep -A 3 1 | wc -l
3
```


runejuhl:
Anything I can do to help get this PR merged? 🙂

koalaman:
Looks good to me, thanks for the bump!

runejuhl:
Thank you for all of your work @koalaman -- shellcheck is invaluable to me!

Issue #2412, opened 2021-12-11
Feature: Warn when function has been defined more than once


Jonta:
#### For new checks and feature suggestions
- [x] https://www.shellcheck.net/ (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/sh

print_greeting () {
        echo "Hi"
}

print_greeting () {
        echo "Hello"
}

print_greeting

```

#### Here's what shellcheck currently says:

"No issues detected"

#### Here's what I wanted or expected to see:

"print_greeting has been defined twice. Perhaps unintentionally"

guillaume-d:
Looks like a duplicate of #2316.

PR #2427, opened 2022-01-01
Add warning for function redeclaration

Flygrounder:
This is a basic implementation of #2412 

However, there are at least two things, that should be discussed:
1. For which line warning should be emitted? For the original declaration, the second one or both?
2. Should warning be emitted for the following script? (Currently it is)
```sh
#!/bin/sh

if [ "$1" = "1" ]
then
print_greeting () {
        echo "Hi"
}
else
print_greeting () {
        echo "Hello"
}
fi

```

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -66,6 +66,7 @@ treeChecks = [
     ,checkUseBeforeDefinition
     ,checkAliasUsedInSameParsingUnit
     ,checkArrayValueUsedAsIndex
+    ,checkFunctionRedeclaration
     ]
 
 runAnalytics :: AnalysisSpec -> [TokenComment]
@@ -4902,6 +4903,15 @@ checkBatsTestDoesNotUseNegation params t =
             [x] -> x == t
             x:rest -> isLastOf t rest
             [] -> False
+
+prop_checkFunctionRedeclaration1 = verifyTree checkFunctionRedeclaration "f () { true; }; f () { true; }"
+prop_checkFunctionRedeclaration2 = verifyNotTree checkFunctionRedeclaration "f () { true; }; g () { true; }"
+
+checkFunctionRedeclaration params t = execWriter $ mapM (\ts -> warn (snd $ ts !! 1) 2216 "Function has been defined twice") $ 
+    filter (\c -> length c > 1) $ groupBy isSameFunc $ sort funcs
+  where
+    funcs = analyse findFunctions t
+    isSameFunc x y = fst x == fst y
 
 return []
 runTests =  $( [| $(forAllProperties) (quickCheckWithResult (stdArgs { maxSuccess = 1 }) ) |])


scottdcoulter:
As one of the reporters of this issue, I would vote for warning on the start line of the second function.
For your second question, I'm of two minds... this use-case looks much more deliberate than the unintentional one I stumbled upon originally, but it might be better to go ahead and flag it, and let the coder put an inline suppression comment if desired.
I have no idea how much work it would be to distinguish this conditional case from a more straightforward/unguarded duplication.
Anyhow, I'm thrilled the issue is being looked into.  Thanks!

src/ShellCheck/Analytics.hs
```diff
@@ -4903,5 +4904,27 @@ checkBatsTestDoesNotUseNegation params t =
             x:rest -> isLastOf t rest
             [] -> False
 
+prop_checkFunctionRedeclaration1 = verifyTree checkFunctionRedeclaration "f () { true; }; f () { true; }"
+prop_checkFunctionRedeclaration2 = verifyNotTree checkFunctionRedeclaration "f () { true; }; g () { true; }"
+
+checkFunctionRedeclaration params t =
+  map
```
koalaman, line 4911: You could simplify this as `execWriter $ mapM (\ts -> warn (snd $ ts !! 1) 2216 "has already been defined twice") $ ..` instead of using `newTokenComment` from scratch

koalaman (commented):
Sorry, I missed this PR!

koalaman:
Ideally it would warn when one definition dominates the other, but there's currently no infrastructure for determining this. I didn't find any such cases in my corpus though, so it's fine if it warns.

koalaman:
I don't know why I didn't see them the first time, but in my more extensive search, it appears that it's significantly more common to intentionally have multiple definitions than to do it accidentally.

Here are the true positives I came across. Interestingly, they're all copypaste failures with identical function bodies:

* https://github.com/tuist/tuist/blob/de1c1d414442c1aaa8b0cd96f948f8d16a53ed98/script/uninstall#L3
* https://github.com/facebook/zstd/blob/fda537b299bfa598e65b47ef77c412dc09313445/tests/gzip/help-version.sh#L226
* https://github.com/crowdsecurity/crowdsec/blob/5f339ab3123dc622dbafae74d7412b514a5b7345/wizard.sh#L45
* https://github.com/netdata/netdata/blob/ce60351a6caee77f433aa4c768d4bc4ed6af0e54/packaging/installer/kickstart.sh#L145

Here are some of the intentional redeclarations, some with different implementations in different branches and some replacing a function for testing:

* https://github.com/bazelbuild/bazel/blob/080d4ad06b07f78fb081ebc483ceb2e3c2bd1be2/src/test/shell/integration/sandboxing_test.sh#L43
* https://github.com/rook/rook/blob/7af10af7816085949c424fd4b9b570e12b4744e5/tests/scripts/dind-cluster.sh#L22
* https://github.com/meteor/meteor/blob/devel/scripts/build-dev-bundle-common.sh#L24
* https://github.com/certbot/certbot/blob/5c111d0bd1206d864d7cb93754e101f6073bc669/letsencrypt-auto-source/letsencrypt-auto#L901
* https://github.com/qutebrowser/qutebrowser/blob/master/misc/userscripts/password_fill#L235
* https://github.com/facebook/hermes/blob/c2cd9e385a922f486a55e6bd70db2032f78379ed/utils/commons.sh#L42
* https://github.com/NixOS/nix/blob/5aeda916390584b7b055ed794d1063a39854f148/scripts/install.in#L85
* https://github.com/corretto/corretto-8/blob/e6eb5c6a8c985ff8f1aa299d87fe5b798b5c2a3e/src/langtools/test/tools/javac/Paths/Util.sh#L33
* https://github.com/facebook/zstd/blob/fda537b299bfa598e65b47ef77c412dc09313445/tests/gzip/init.sh#L337

I tried restricting the check to only warn about top level definitions, but these were also more intentional than not:
* https://github.com/tteck/Proxmox/blob/c14b0673ed42f35e3e94fd747fed1943bdb025ef/ct/debian-v2.sh#L40
* https://github.com/returntocorp/semgrep/blob/e8bcf4248f43543730763a4a22bdde95f698513c/semgrep-core/src/spacegrep/scripts/show-perf#L90
* https://github.com/nvm-sh/nvm/blob/7b530edd3a2a3ebc976c455394f7245827194da7/test/fast/Unit%20tests/nvm_print_implicit_alias%20success#L50
* https://github.com/niieani/bash-oo-framework/blob/0a9ea03e74256f3e4eec252623126b801d3491e2/example/v1/logging.sh#L27

Unfortunately it does not appear that this check will provide the signal-to-noise ratio we expected it would. I appreciate the contribution though!

Flygrounder:
Fair enough. Thanks for taking time to test and review this!

PR #2494, opened 2022-05-05
add SC2316: error on multiple declarations like 'readonly local'

patrickxia:
As it says on the tin. This is a common mistake that I see all over the place.

Note that you might see an "unused variable" warning for some cases of this, but (a) it's easy to accidentally ignore and (b) this is not the case for the "export readonly" variant of this mistake.


koalaman:
Very nice, thanks!  I switched out `getLiteralString` with `getUnquotedLiteral` to allow you to clarify with quotes as in `readonly "local"`, and added a wiki page.

PR #2500, opened 2022-05-16
ci: update github actions

Fdawgs:
This PR updates GitHub Actions to their latest major versions. Main change is they are all now using Node 16 internally instead of 12, which was EOL at the end of April.

- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Releases for `actions/upload-artifact` can be [found here](https://github.com/actions/upload-artifact/releases)
- Releases for `actions/download-artifact` can be [found here](https://github.com/actions/download-artifact/releases)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
           sudo apt-get install cabal-install
 
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


koalaman:
Thanks!

nelsonjchen:
FWIW: It's also possible to teach dependabot to do and propose this. 

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions

PR #2527, opened 2022-07-07
Allow 'su --command=something'

pSub:
Previously, 'su --command=something' would result in SC2117 although
it is the official long version of 'su -c something'.

### DIFF
--- a/src/ShellCheck/Checks/Commands.hs
+++ b/src/ShellCheck/Checks/Commands.hs
@@ -607,9 +607,11 @@ prop_checkInteractiveSu1 = verify checkInteractiveSu "su; rm file; su $USER"
 prop_checkInteractiveSu2 = verify checkInteractiveSu "su foo; something; exit"
 prop_checkInteractiveSu3 = verifyNot checkInteractiveSu "echo rm | su foo"
 prop_checkInteractiveSu4 = verifyNot checkInteractiveSu "su root < script"
+prop_checkInteractiveSu5 = verify checkInteractiveSu "su -c something"
+prop_checkInteractiveSu6 = verify checkInteractiveSu "su --command=something"
 checkInteractiveSu = CommandCheck (Basename "su") f
   where
-    f cmd = when (length (arguments cmd) <= 1) $ do
+    f cmd = when (noCommandArgPresent (arguments cmd)) $ do
         path <- getPathM cmd
         when (all undirected path) $
             info (getId cmd) 2117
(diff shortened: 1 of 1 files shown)


src/ShellCheck/Checks/Commands.hs
```diff
@@ -607,9 +607,11 @@ prop_checkInteractiveSu1 = verify checkInteractiveSu "su; rm file; su $USER"
 prop_checkInteractiveSu2 = verify checkInteractiveSu "su foo; something; exit"
 prop_checkInteractiveSu3 = verifyNot checkInteractiveSu "echo rm | su foo"
 prop_checkInteractiveSu4 = verifyNot checkInteractiveSu "su root < script"
+prop_checkInteractiveSu5 = verify checkInteractiveSu "su -c something"
```
koalaman, line 610: This unit test makes sure you *do* get a warning :P 

You can use `verifyNot` to make sure you don't

src/ShellCheck/Checks/Commands.hs
```diff
@@ -620,6 +622,9 @@ checkInteractiveSu = CommandCheck (Basename "su") f
     undirected (T_Redirecting _ (_:_) _) = False
     undirected _ = True
 
+    noCommandArgPresent = not . any isCommandArg . map (getLiteralStringDef "_")
+    isCommandArg arg = arg == "-c" && "--command=" `isPrefixOf` arg
```
koalaman, line 626: This is always false because of the `&&`, and would fail to recognize `--command foo` (without `=`) since it would now apply to multiple arguments as well, and something like `su -Pc cmd` is valid if rare.

src/ShellCheck/Checks/Commands.hs
```diff
@@ -620,6 +622,9 @@ checkInteractiveSu = CommandCheck (Basename "su") f
     undirected (T_Redirecting _ (_:_) _) = False
     undirected _ = True
 
+    noCommandArgPresent = not . any isCommandArg . map (getLiteralStringDef "_")
```
koalaman, line 625: There are some other valid ways of specifying a command, such as `su root ./myscript`, while noobs more frequently just use `su`, `su -`, or `su root` when encountering this pitfall

koalaman (changes requested):
Neat! For the reasons mentioned I think you could fix the bug while also not introducing false positives by keeping the logic where it checks for 0 or 1 arguments (as this is the most common way people encounter the pitfall), and adding the additional requirement that when there's a single argument it does not start with `--command=`. Maybe along the lines of

```
    unless (commandSpecified $ arguments cmd) $ do ..
  where
    commandSpecified [] = False
    commandSpecified [arg] = "--command=" `isPrefixOf` arg
    commandSpecified _ = True
```

PR #2535, opened 2022-07-17
SC2183 grammer fix: 'variable' instead of 'variables' if only one variable

### LINKED ISSUE #2400: SC2183: Grammar: "1 variables" instead of "1 variable"
- SC2183
- 0.7.1

```sh

#!/bin/sh
printf "%s\n"

```

       ^-- SC2183 (warning): This format string has 1 variables, but is passed 0 arguments.

       ^-- SC2183 (warning): This format string has 1 variable, but is passed 0 arguments.

ygeyzel:
fix #2400


koalaman:
I additionally applied the same logic to "arguments" in  ea4e0091c7b4. Thanks!

Issue #2537, opened 2022-07-18
SC2311 shouldn't appear with set -o posix


ale5000-git:
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2311
- My shellcheck version (`shellcheck --version` or "online"):  online
- [x] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2086)
- [x] I tried on https://www.shellcheck.net/ and verified that this is still a problem on the latest commit

#### For new checks and feature suggestions
- [x] https://www.shellcheck.net/ (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```bash
#!/usr/bin/env bash
# shellcheck enable=all

set -e
set -o posix

test_func()
{
  :
}

_test="$(test_func)"
```

#### Here's what shellcheck currently says:
[SC2311](https://www.shellcheck.net/wiki/SC2311) (info): Bash implicitly disabled set -e for this function invocation because it's inside a command substitution. Add set -e; before it or enable inherit_errexit.

#### Here's what I wanted or expected to see:
`set -o posix` has the effect to also enable `inherit_errexit`, so SC2311 shouldn't appear

koalaman:
Fixed by @DoxasticFox, thanks!

ale5000-git:
@koalaman: It is still not fixed here => https://www.shellcheck.net/

ale5000-git:
Now it is fixed, thanks.

PR #2586, opened 2022-09-23
Suppress SC2311 with `set -o posix`

DoxasticFox:
Resolves https://github.com/koalaman/shellcheck/issues/2537.


koalaman:
Thanks!

Issue #2563, opened 2022-08-16
False positive SC2094


Bilge:
- SC2094
- [x] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2094)
- [x] I tried on https://www.shellcheck.net/ and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/bash
mapfile -t foo <foo
```

#### Here's what shellcheck currently says:
Make sure not to read and write the same file in the same pipeline.
See SC2094.

#### Here's what I wanted or expected to see:
Nothing 🙂. `mapfile` creates a variable, not a file.

kurahaupo:
Agreed, that's a bug, and it's apparent on [shellcheck.net](https://shellcheck.net).

koalaman:
Fixed by @DoxasticFox, thanks!

PR #2588, opened 2022-09-24
Add `mapfile` to harmless commands for SC2094

DoxasticFox:
Resolves https://github.com/koalaman/shellcheck/issues/2563.


koalaman:
Thanks!

PR #2601, opened 2022-10-07
Permit colon after exec

mavit:
ShellCheck throws warning SC2093 when a script contains commands that could never be executed because they are after an `exec`.  Command `:` does nothing, so add it to the list of commands that don’t trigger this warning.


mavit:
> `:` after a failing exec would reset the failing exit status.

Ah, good point.  That doesn't affect me because I'm using `set -o errexit`, but it would be an issue if I wasn't.

Perhaps, then, `:` should be ignored only if `errexit` is in effect.  I would have a go at updating the pull request to do that, but I'm afraid I don't know any Haskell.

> Did this come up in a real world scenario?

Here’s the scenario.  I have a codebase that’s a mixture of Perl and shell scripts.  The Perl is documented using its built-in markup format, [POD](https://perldoc.perl.org/perlpod), and it’s convenient to use the same for the shell scripts.  The trick for embedding POD into a shell script without confusing the shell is to use a here-document with a delimiter that is the same as POD's delimiter:

```bash
: <<'=cut'

=head1 DESCRIPTION

Here would be some documentation.

=cut
```

A natural place to want to put this is at the end of the script.

koalaman:
Huh, I've never seen this before. However, I do see a small handful of scripts using this, and zero false positives. Thanks!

Issue #2550, opened 2022-08-05
SC2312: false positive when using 'time'


dtmdl:
#### For bugs
- Rule Id: SC2312
- My shellcheck version: 0.8.0 and online
- [X] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2086)
- [X] I tried on https://www.shellcheck.net/ and verified that this is still a problem on the latest commit

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/bash
# shellcheck enable=all

set -ex

echo before
time false
echo after

```

#### Here's what shellcheck currently says:
```
[Line 7:](javascript:setPosition(7, 6))
time false
     ^-- [SC2312](https://www.shellcheck.net/wiki/SC2312) (info): Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore).
```

#### Here's what I wanted or expected to see:
No warning, as the exit code is not masked. The script will abort if run:

```
$ ./t.sh || echo 'failed'
+ echo before
before
+ false

real    0m0.000s
user    0m0.000s
sys     0m0.000s
failed
```

PR #2602, opened 2022-10-08
Fix false positive for SC2312 when using `time`

DoxasticFox:
Resolves https://github.com/koalaman/shellcheck/issues/2550.

@koalaman are there any commands in addition to `time` whose parsed arguments might have the `T_SimpleCommand` type? I expected `watch` and `eval` to have the same issue as `time`, but they're fine.


DoxasticFox:
Actually this PR is wrong. It produces no error for the following example when it should:

```bash
#!/bin/bash
# shellcheck enable=all

x=$(false) time false
```

DoxasticFox:
Alright, I think it's ready for review now :crossed_fingers:

koalaman:
Nice, LGTM! `time` is a bit of an odd duck because in bash it's shell syntax and not a command name. You can see this in how `time for((;;)); do true; done` is valid, but with watch/eval it's a syntax error.

PR #2625, opened 2022-11-17
Warn about 'read' without a variable in POSIX sh

ArenM:
According to the posix specification read takes one or more variable names. Dash throws an error if the read command isn't supplied a variable name.


koalaman:
Thanks! I additionally fixed a crash for `read ""` in 46b678fca8f, and moved the snippet to a more general definition of `bashism` since only one definition ends up running and this allows `read -e` to trigger both this new warning and about `-e` being unsupported.

PR #2640, opened 2022-12-15
Add Guix install and usage instructions

jgarte:


port19x:
I suppose it would only be fair.
Tho I find it strange that distro specific install instructions are provided for the distros anyway.
You would assume that developers are capable of installing a package.
A better way to notify developers of native package availability would be a repology badge.
[Packaging status](https://repology.org/project/shellcheck/versions)

Tho that would be a different PR.

port19x (approved):

PR #2683, opened 2023-02-04
Support fgl 5.8.1.0.

### LINKED ISSUE #2677: prop_checkOverwrittenExitCode8 fails with fgl-5.8.1.0
#### For bugs
- My shellcheck version (`shellcheck --version` or "online"): 0.9.0

With fgl upgraded from 5.8.0.0 to 5.8.1.0, the test "prop_checkOverwrittenExitCode8" now fails with an exception:
```
=== prop_checkOverwrittenExitCode8 from src/ShellCheck/Analytics.hs:4927 ===
*** Failed! Exception: '(Array.!): undefined array element' (after 1 test):
```

All other dependencies remained at the same versions during the test.

The only change in fgl seems to be https://github.com/haskell/fgl/commit/e26f7bb36642bc52183a382485ffe89dbe526569 and I don't really understand it.

athas:
ShellCheck used to rely on a bug in
Data.Graph.Inductive.Query.Dominators.dom where a node not reachable from the root would be considered dominated by every node in the graph.

This commit restores that functionality, and is therefore compatible with both fgl 5.8.1.0 and previous versions.

Fixes #2677.


koalaman:
Thanks for debugging this! You're right about the root cause of this failing test, but unfortunately the fgl diff that changed this behavior also introduced a crash bug, so 5.8.1.0 still isn't compatible. I filed https://github.com/haskell/fgl/issues/109

athas:
Is there some expanded test suite I can use to check whether my fix will work?  `cabal test` worked for me, but you seemed to find this other error pretty quickly.

koalaman:
There's no formal extended test suite, but I like to diff the ShellCheck output on a few thousand random scripts. In this case I happened to have a `/usr/src/linux-headers-5.15.0-46/tools/testing/selftests/net/icmp_redirect.sh` from Ubuntu's linux-headers package that triggered the bug.

I also have a corpus of shell scripts scraped from GitHub, but didn't track the provenance. At some point I should try to scrape a set of BSD/GPL licensed ones and that I can put in a public repo as test data.

PR #2808, opened 2023-07-28
improve short description for SC2038

bruce-ricard:
* The short description used to say that until commit aac7d76047a5b28d064b17a5d0fac022054d05a0 from 2014. It appears that it was changed by mistake in that commit to something less readable.

* With the message "use -print0/-0" we were confused and introduced a bug in our code because we didn't understand what to do with the "-0".

* SC2011 (source https://github.com/koalaman/shellcheck/blob/c9e27c24700cdc5b84cfca1f7a90fe07f542867c/src/ShellCheck/Analytics.hs#L591) uses that exact warning message, we copied it from there.

cc: @swalchemist


koalaman:
Thanks!

PR #2825, opened 2023-09-05
Drop root privileges in Docker container

jtesta:
By default, Docker runs containers with root privileges (!).  This isn't necessary for shellcheck.  This PR causes the container to be run as an unprivileged user instead.

FYI, the highest possible UID and GID (65535) must be used in this patch since the final scratch image does not include /etc/passwd, /etc/group, nor the support code to resolve names to UIDs/GIDs.

### DIFF
--- a/Dockerfile.multi-arch
+++ b/Dockerfile.multi-arch
@@ -23,4 +23,8 @@ FROM scratch
 LABEL maintainer="Vidar Holen <vidar@vidarholen.net>"
 WORKDIR /mnt
 COPY --from=alpine /bin/shellcheck /bin/
+
+# Drop root privileges.
+USER 65535:65535
+
 ENTRYPOINT ["/bin/shellcheck"]


koalaman:
This change risks breaking CI and requires workarounds for anyone who's not checking world-readable files. Are there any Docker guidelines or conventions that recommend this approach?

jtesta:
The Center for Internet Security (CIS) Benchmark for Docker states in section 4.1 that containers should be run as non-root whenever possible (see https://www.cisecurity.org/benchmark/docker).  Furthermore, running as non-root by default would be applying the Principle of Least Privilege.

As for filesystem permissions, default Ubuntu systems have a umask of 0002 (meaning files are already world-readable).  So this would not be a problem.  In the event that this is changed, though, users can add `-u $(id -u):$(id -g)` to their `docker run` command, which would run the container as the host user.

Because most users simply copy/paste from the documentation, we can very easily add the `-u` part to https://github.com/koalaman/shellcheck/blob/master/README.md?plain=1#L214.  Adding `-u` to the CI config would also be easy.

ZetaMaster777 (approved):

meadowbees:
Running as an unprivileged user makes sense for reducing security risks. Good call! I agree with this.

jtesta:
Bump.

e-kwsm (approved):

PR #2876, opened 2023-12-06
Testsuite: report which module failed the tests

andreasabel:
This also fixes the problem that the testsuite threw `exitFailure`
even when it succeeded (which I found inexplicable).

Once this is published, the testsuite could be enabled in Stackage again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ cabal.config
 /parts/
 /prime/
 *.snap
+/dist-newstyle/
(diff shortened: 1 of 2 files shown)


andreasabel:
CI does not run here, maybe add `pull_request` as trigger? https://github.com/koalaman/shellcheck/blob/a71a13c2fca05b822cb16840792dc013ca76493f/.github/workflows/build.yml#L3-L5

grische:
> CI does not run here, maybe add pull_request as trigger?

@andreasabel that is currently not possible. One would need to rewrite the action to allow running it on PRs.

koalaman:
A nice quality-of-life improvement. Thanks!

PR #2879, opened 2023-12-14
Add installation directions for winget

### LINKED ISSUE #2691: winget
Hello,

I am a Windows user wanting to use Shellcheck. In the readme, I see that it says that I should use `chocolatey` or `scoop` to install Shellcheck.

However, the standard way to install packages on Windows is with [`winget`](https://learn.microsoft.com/en-us/windows/package-manager/winget/), which is automatically installed on all Windows computers without having to do anything. For example, to install `git`, all you have to do on a modern Windows computer is to type: `winget install git`

For this reason, `chocolatey` and `scoop` should be considered deprecated.

I tried to run `winget install shellcheck`, but it didn't work, so it looks like you guys don't have a Windows package. Please consider creating one for your Windows users.

slycordinator:
ShellCheck is now available on winget, so we can add it to the installation methods.

Fixes #2691

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,12 @@ On Windows (via [chocolatey](https://chocolatey.org/packages/shellcheck)):
 C:\> choco install shellcheck
 ```
 
+Or Windows (via [winget](https://github.com/microsoft/winget-pkgs)):
+
+```cmd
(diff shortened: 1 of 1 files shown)


koalaman:
Thanks!

PR #2885, opened 2023-12-15
.cabal: allow Diff-0.5

juhp:
see also https://github.com/commercialhaskell/stackage/issues/7159


juhp:
I confirmed locally that it builds fine with Diff-0.5

koalaman:
Thanks!

Issue #2199, opened 2021-04-06
[Feature] Pass path to shellcheckrc from command line


mohkale:
#### For new checks and feature suggestions
- [X] https://www.shellcheck.net/ (i.e. the latest commit) currently gives no useful warnings about this
- [X] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Description
At the moment there doesn't seem to be a way to tell shellcheck to read a specific configuration file instead of looking for the global one. 

For example I've got a [lint](https://github.com/mohkale/dotfiles/blob/cdd4ecaeb6c1fbc6e6390035c79d55a6e7a202aa/setup/lint) script on my dotfiles and I'd like to setup a github action that uses my shellcheckrc file at `core/scrc`, how would you do this?

#### Version Info
```
ShellCheck - shell script analysis tool
version: 0.7.1
license: GNU General Public License, version 3
website: https://www.shellcheck.net
```

koalaman:
That would make sense. ShellCheck does already look at the script's closest parent directory with a `.shellcheckrc`, but that doesn't seem to help in your case.

cavo789:
Hello

Same need for me: I've a set of hooks that will verify the quality of my files before pushing them to the remote server. In my particular use case, I need to be able to specify where the `.shellcheckrc` should be retrieved like, f.i. `shellcheck --config=.config/.shellcheckrc` (I'm putting all my configurations files there).

The parameter can be, of course whatever you wish, most of static analysis tools used `--config`.

Thanks!

diederikdehaas:
I'd very much like this option too (for CI purposes). I hope it's not too greedy to ask to have this option included in [the next release](https://github.com/koalaman/shellcheck/issues/2598) so that it could become part of the next Debian Stable (Bookworm) release.

Cheers! And thank you for this wonderful tool :-)

TheCaptain989:
Just adding my name to the list for this feature.  It would be particularly useful when Shellcheck is called from Super-Linter.

PR #2908, opened 2024-01-21
Implement rcfile option

tacerus:
This introduces the "--rcfile" argument which allows a specific shellcheckrc file to be passed.
If specified and the given file exists, the default locations will not be searched and the specified file will be used.


koalaman:
Thanks! I added some minor modifications in 6a44a19f17c4 to disable normal search when the file is unavailable, and to cache the contents as with other files.

tacerus:
Thank you! This was my first Haskell patch, good to see your improvements. :-)

I think this resolves https://github.com/koalaman/shellcheck/issues/2199.

PR #2917, opened 2024-02-02
Remove deprecated "install --enable-tests" command

grische:
This command does not work anymore and it seems that a `cabal test` is sufficient without an install beforehand.


koalaman:
Thanks!

PR #2938, opened 2024-03-08
Recommend `typeset` instead of `declare` in SC2324

larryv:
Bash has both `typeset` and `declare`, but ksh has `typeset` only. Recommend the more portable alternative to users.


koalaman:
Thanks!

PR #2960, opened 2024-04-04
Update README.md to add macOS Arm64 pre-compiled binaries link

hugos99:
Currently README file doesn't include the macOS Arm64 pre-compiled binaries link.

This PR just adds that link for clarity.


koalaman:
Thanks!

PR #2973, opened 2024-04-25
Add new bats variables stderr and stderr_lines

jandubois:
These are being set by `run --separate-stderr` and have been introduced in https://github.com/bats-core/bats-core/releases/tag/v1.5.0


koalaman:
Thanks!

PR #2992, opened 2024-05-30
Add linux.riscv64 precompiled support

mengzhuo:


koalaman:
Ah welp, it looked good to me but it turned out to accidentally rely on binfmt_misc instead of QEMU_EXECVE, which the Github runner doesn't have. It should be a simple fix by switching to the QEmu fork that has QEMU_EXECVE and doing the three hour rebuild. I'll get that going soon.

mengzhuo:
Thanks!

koalaman:
I don't know why I had the hubris to say it should be a simple fix.

`echo 0 | sudo tee /proc/sys/fs/binfmt_misc/status` disables binfmt_misc on the host to allow reproducing the GitHub Actions failure.

I spent a ridiculous amount of time kludging and debugging only to discover that `hashable` was updated last week, and now fails on riscv64 due to its lack of support for `-march=native`. My docker cache was full of images from before this change, making debugging especially fun.

The problem can apparently be worked around with `cabal install .. -c 'hashable -arch-native'`, but I still can't find that mentioned anywhere. The documented answer is `-f -arch-native`, which works to build the individual package, but does not allow it to be reused as a dependency by ShellCheck. 

https://github.com/balena-io/qemu where QEMU_EXECVE originates appears to have a memory corruption issue interpreting shebangs, but the older fork https://github.com/koalaman/qemu used by the armv6hf image mostly works. Some binaries segfault at random though, which I hoped a later version would fix. I've added some retries and will try updating later. 

The memory usage of GHC was due to QEmu's overhead per vspace, which I fixed by setting an ulimit to avoid the GHC runtime from allocating it's 1TB default. It no longer needs `--jobs=1`. To further speed up the build, I kludged it to use a cabal freeze file to install dependencies and to ensure they get used, so now it builds in 30 minutes on my machine.

Anyways, it works and is now merged (Thanks!), but the same `hashable` change has broken some of the other archs, so it's not quite ready to test yet.

koalaman:
The riscv64 binaries are now being published on https://github.com/koalaman/shellcheck/releases/download/latest/shellcheck-latest.linux.riscv64.tar.xz 

Do you have some real hardware to test it on? If it works I can backport and upload a stable version.

mengzhuo:
> The riscv64 binaries are now being published on https://github.com/koalaman/shellcheck/releases/download/latest/shellcheck-latest.linux.riscv64.tar.xz
> 
> Do you have some real hardware to test it on? If it works I can backport and upload a stable version.

Thanks!
I'd tested on my riscv box and it works as expected!

PR #3011, opened 2024-06-28
Fix SC3003, SC3036 and SC3045 for busybox shell

sertonix:
Busybox shell supports `$'..'`, `read -p <prompt>`, `type -p` and `echo -e` so remove the warnings there.


koalaman:
Nice, thanks!

PR #3017, opened 2024-07-08
Add a reason directive to aid in documenting annotations

### LINKED ISSUE #3016: Add a reason directive to document why some checks were disabled
- [ ] https://www.shellcheck.net/ (i.e. the latest commit) currently gives no useful warnings about this

```sh
#!/bin/bash

function blah {
  echo hello $1
}

# shellcheck disable=SC2044 - we call a local function which is not available when using find with -exec
for archive in $(find "$targetdir" -name '*.tar.gz'); do
  blah "$archive"
done
```

```
In shared/log4shell.sh line 113:
  # shellcheck disable=SC2044 - we call a local function which is not available when using find with -exec
  ^-- SC1073 (error): Couldn't parse this shellcheck directive. Fix to allow more checks.
                               ^-- SC1072 (error): Expected '=' after directive key. Fix any mentioned problems and try again.
```

I would like there to be a `reason` directive so we can document why we disable some checks. Eg:

```sh
# shellcheck disable=SC2044 reason="we call a local function which is not available when using find with -exec"
```

NickLarsenNZ:
Fixes #3016

> [!NOTE]
> This is not yet tested locally, but I wanted to raise it to get any other feedback on the contribution.

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -1083,6 +1083,10 @@ readAnnotationWithoutPrefix sandboxed = do
                         parseNoteAt pos ErrorC 1145 "Unknown external-sources value. Expected true/false."
                         return []
 
+            -- noop to allow a reason to be given for the annotation
+            "reason" -> do
+                return []
+
             _ -> do
                 parseNoteAt keyPos WarningC 1107 "This directive is unknown. It will be ignored."
                 anyChar `reluctantlyTill` whitespace


NickLarsenNZ:
Closing because I have found a way to add comments inline based on the props:

https://github.com/koalaman/shellcheck/blob/98b8dc0720148d69ff924f89966c50dc2dda2fe3/src/ShellCheck/Parser.hs#L991-L992

NickLarsenNZ:
Closing, as this is possible, just not documented from what I can see.

For anyone trying to do the same, you can do it like so:

```sh
shellcheck disable=SC1234 # here goes the reason
```

PR #3106, opened 2024-12-28
Recognize internal variables new in bash 5.3

larryv:
From the bug-bash@gnu.org announcement "[Bash-5.3-beta available](https://lists.gnu.org/archive/html/bug-bash/2024-12/msg00120.html)":

> q. GLOBSORT: new variable to specify how to sort the results of pathname expansion (name, size, blocks, mtime, atime, ctime, none) in ascending or descending order.
> 
> w. BASH_MONOSECONDS: new dynamic variable that returns the value of the system's monotonic clock, if one is available.
> 
> x. BASH_TRAPSIG: new variable, set to the numeric signal number of the trap being executed while it's running.


koalaman:
Thanks!

PR #3108, opened 2024-12-31
Add bang as a valid char for function names

### LINKED ISSUE #3102: Joined pound in function name parse failed
#### For bugs
- Rule Id: SC1073
- My shellcheck version (`shellcheck --version` or "online"):

```sh
#!/usr/bin/bash

function x#y {
    echo foo
}

x#y
```

```

In - line 3:
function x#y {
^-- SC1073 (error): Couldn't parse this function. Fix to allow more checks.


In - line 4:
    echo foo
    ^-- SC1064 (error): Expected a { to open the function definition.
    ^-- SC1072 (error):  Fix any mentioned problems and try again.

For more information:
  https://www.shellcheck.net/wiki/SC1064 -- Expected a { to open the function...
  https://www.shellcheck.net/wiki/SC1072 --  Fix any mentioned problems and t...
  https://www.shellcheck.net/wiki/SC1073 -- Couldn't parse this function. Fix...
```

no outputs

Flu:
As described by issue #3102, the parser fails when a function name contains a pound symbol. As far as I can tell, it is a valid way of naming functions in bash, it works with the `function` keyword and without. Not sure how portable it is, but I can add a check for that as part of this PR if it's desirable by @koalaman 

The only thing I can find about function names in bash is here:
https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions

It only explicitly forbids `$` in function names.

This closes #3102

### DIFF
--- a/src/ShellCheck/Parser.hs
+++ b/src/ShellCheck/Parser.hs
@@ -67,7 +67,7 @@ doubleQuote = char '"'
 variableStart = upper <|> lower <|> oneOf "_"
 variableChars = upper <|> lower <|> digit <|> oneOf "_"
 -- Chars to allow in function names
-functionChars = variableChars <|> oneOf ":+?-./^@,"
+functionChars = variableChars <|> oneOf "#:+?-./^@,"
 -- Chars to allow in functions using the 'function' keyword
 extendedFunctionChars = functionChars <|> oneOf "[]*=!"
 specialVariable = oneOf (concat specialVariables)


koalaman:
Thanks! ShellCheck has historically been slightly restrictive about special characters in function names without `function` because it tended to result in confusing parsing errors when given common newbie syntax mistakes, but this addition had no such negative impact.

PR #3114, opened 2025-01-07
Skip SC2015 when the last command is true

### LINKED ISSUE #2977: Skip SC2015 when it end with: || true
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2015

```sh
#!/bin/sh
set -e

foo && bar || true
```

```
foo && bar || true
    ^-- [SC2015](https://www.shellcheck.net/wiki/SC2015) (info): Note that A && B || C is not if-then-else. C may run when A is true.
```

Nothing, the `|| true` is used to ignore return codes so even if it get executed when A is true it doesn't matter.

Flu:
In cases like 
```bash
#!/bin/sh
set -e
foo && bar || true
```
The last `true` is used for masking the return value, no matter if both `foo` and `bar` executed, therefore SC2015 has no relevance here.

Closes #2977

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -881,13 +881,15 @@ prop_checkShorthandIf6 = verifyNot checkShorthandIf "if foo && bar || baz; then 
 prop_checkShorthandIf7 = verifyNot checkShorthandIf "while foo && bar || baz; do true; done"
 prop_checkShorthandIf8 = verify checkShorthandIf "if true; then foo && bar || baz; fi"
 prop_checkShorthandIf9 = verifyNot checkShorthandIf "foo && [ -x /file ] || bar"
+prop_checkShorthandIf10 = verifyNot checkShorthandIf "foo && bar || true"
+prop_checkShorthandIf11 = verify checkShorthandIf "foo && bar || false"
 checkShorthandIf params x@(T_OrIf _ (T_AndIf id _ b) (T_Pipeline _ _ t))
         | not (isOk t || inCondition) && not (isTestCommand b) =
     info id 2015 "Note that A && B || C is not if-then-else. C may run when A is true."
   where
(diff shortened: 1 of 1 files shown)


SvenStaehs:
Thank you! Hope to see this included in the next release, we are due one any day now 😆 
My use-case for the record:
```sh
for attempt in {3..0}; do
    [[ $attempt -gt 0 ]] || die "Failed to [do the thing]."
    # shellcheck disable=SC2015 # "break" never fails (I think), AND "true" has no side-effect, so this rule does not apply. See https://github.com/koalaman/shellcheck/issues/2977
    doSomethingFlaky && break || true
done
```

src/ShellCheck/Analytics.hs
```diff
@@ -881,13 +881,15 @@ prop_checkShorthandIf6 = verifyNot checkShorthandIf "if foo && bar || baz; then
...
+prop_checkShorthandIf10 = verifyNot checkShorthandIf "foo && bar || true"
+prop_checkShorthandIf11 = verify checkShorthandIf "foo && bar || false"
 checkShorthandIf params x@(T_OrIf _ (T_AndIf id _ b) (T_Pipeline _ _ t))
         | not (isOk t || inCondition) && not (isTestCommand b) =
     info id 2015 "Note that A && B || C is not if-then-else. C may run when A is true."
   where
     isOk [t] = isAssignment t || fromMaybe False (do
         name <- getCommandBasename t
-        return $ name `elem` ["echo", "exit", "return", "printf"])
+        return $ name `elem` ["echo", "exit", "return", "printf", "true"])
```
e-kwsm, line 892: colon(1P), `:`, does nothing and always returns zero as well.

```suggestion
        return $ name `elem` ["echo", "exit", "return", "printf", "true", ":"])
```

koalaman:
Merged with the addition of `:` as suggested. Thanks!

PR #3115, opened 2025-01-07
Add optional check for whether a variable can be `readonly`

### LINKED ISSUE #2987: [Feature request][Bash] Add rules to provide tips on adding local, or readonly status to variables
Hello :),

I may have used readonly or local keywords in some of my shell scripts in the past.
But at some point, I forgot about this keywords.

See https://github.com/koalaman/shellcheck/issues/1944
for the issue that recalled me of it.
I checked that no current rule (activated with --enable=all for example) yields suggestions to add such keywords in your code.
I have a very simple rule to propose:
If there is no match of the regexp like "^\s*@my_variable_name@=.*$" adapted to the variable in the current file, propose to add readonly keyword. Another feature would be to define a project variant of this to check all lines of bash scripts in . for example instead of just the current file. Regarding local vs global variable in functions, it could make sense that each variable in a function is defined with local or declare or typeset or that a warning is issued by some rule (the case of exterior variables could be dealed with `# shellcheck disable=SC1111(@my_variable_name@)` at beginning of function with function scope.

Best regards,
     Laurent Lyaudet

Flu:
Adds an optional check for triggering a SC2331 for variables that are declared (and/or initialized) but never assigned to again. If a variable is found that matches, a note is given to maybe declare it as `readonly`.

Also implemented tests for it, as well as for the option itself in Checker.hs

Closes #2987

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -71,7 +71,6 @@ treeChecks = [
     ,checkUseBeforeDefinition
     ,checkAliasUsedInSameParsingUnit
     ,checkArrayValueUsedAsIndex
-    ,checkVariableCanBeReadonly
     ]
 
 checker spec params = mkChecker spec params treeChecks
@@ -280,6 +279,13 @@ optionalTreeChecks = [
         cdPositive = "cat foo | grep bar",
         cdNegative = "grep bar foo"
     }, nodeChecksToTreeCheck [checkUuoc])
+
+    ,(newCheckDescription {
+        cdName = "check-variable-can-be-readonly",
+        cdDescription = "Check that a variable can be made readonly if it isn't assigned to.",
+        cdPositive = "x=3; echo $x",
+        cdNegative = "readonly x=3; echo $x"
+    }, checkVariableCanBeReadonly)
     ]
 
 optionalCheckMap :: Map.Map String (Parameters -> Token -> [TokenComment])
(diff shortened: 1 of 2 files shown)


koalaman:
This a neat check, but unfortunately ShellCheck's poor DFA makes it a lot trickier than it ought to be:

* It doesn't handle loops, e.g. `for file in *; do name=${f%.*}; echo "$name.txt"; done` will trigger for both `file` and `name`
* Similarly, functions like `f() { last=$1; }; f bar; echo $last; f baz; echo $last;`
* `[ -z "$file" ]` triggers for `$file` due to the original DFA algorithm treating this as a signal that the user has checked the variable for existence (awful hack on my part)

ShellCheck does have a significantly more correct DFA algorithm that does track control flow properly (ShellCheck.CFG), but it is correspondingly more complex and would require a whole lot more compiler type graph operations in order to determine that the variable is only set once.

Flu:
> ShellCheck does have a significantly more correct DFA algorithm that does track control flow properly (ShellCheck.CFG), but it is correspondingly more complex and would require a whole lot more compiler type graph operations in order to determine that the variable is only set once.

So does this mean that using ShellCheck.CFG would make this kind of check possible, but it would be complicated to take into account all of the possible edge cases? Cause if so, I could still try. I don't know if it's worth the effort for such a small and optional check, but I could give it a shot if it's technically possible

PR #3175, opened 2025-03-24
fix(SC3013): check POSIX-compliant unary operators for test and [

### LINKED ISSUE #2125: Add warnings for `sh` for non-POSIX extensions

This is just a small example, please read **More details** section below.

```sh
#!/bin/sh 
[ -k foo ]
```

No issues detected!

In POSIX sh, -k is undefined.

(Or something similar.)

#### More details

Shellcheck does not emit warnings for the non-POSIX operators `-nt`, `-ef`, `-k`, `-o`, `<`, and `>` when using a `#!/bin/sh` shebang or `--shell sh`.

Shellcheck already emits warnings for non-POSIX syntax ([example 1](https://github.com/koalaman/shellcheck/wiki/SC2039), [example 2](https://github.com/koalaman/shellcheck/wiki/SC3026)). It would be useful if shellcheck could also warn for these non-POSIX operators, when the `sh` shell is specified.

e-kwsm:
fix #2125


koalaman:
Thanks! I refactored a bit in dc41f0cc5bdcf1c8 to make adding future flags less annoying.

PR #3223, opened 2025-06-18
snap: switch to `core24` base

simondeziel:


simondeziel:
@koalaman while I've smoked tested the resulting snap, it's my first experience with haskell and the cabal build system so please bear that in mind while reviewing ;)

If you have a Launchpad account, you can easily launch a remote build with: `snapcraft remote-build --launchpad-accept-public-upload --build-for amd64`

koalaman:
This works for amd64, thanks!

Several other architectures were failing to build for unknown reasons both before and after, so I'll leave it merged.

Issue #2689, opened 2023-02-08
Inconsistent triggering of SC2268 (x-prefix comparisons)


slycordinator:
SC2268 recommends that one not use x-prefix comparisons, but it allows them in many cases.

SC2268 appears to only get triggered if:
1) The comparison is = or ==. Nothing happens when the comparison is '!="
[ ! "x$var" = "x" ] and [ ! "x$var" == "x" ] trigger the warning, but [ "x$var" != "x" ] and [ ! "x$var" != "x" ] do not trigger it.

2) The prefix is solely the lowercase 'x'. Any other prefix doesn't trigger the issue. Recently came across scripts that occasionally have  "X" as the prefix and it doesn't get triggered. Also doesn't get triggered when any other letter is used.
[ "x$var" = "x" ] triggers this, but [ "X$var" = "X" ] does not.

Thank you

#### For bugs
- Rule Id 2068: 
- My shellcheck version (`shellcheck --version` or "online"): 
- [x] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2086)
- [X] I tried on https://www.shellcheck.net/ and verified that this is still a problem on the latest commit

#### For new checks and feature suggestions
- [x] https://www.shellcheck.net/ (i.e. the latest commit) currently gives no useful warnings about this
- [X] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

#### Here's a snippet or screenshot that shows the problem:

```sh

#!/bin/bash
#shellcheck disable=SC2154

if [ "x$var" = "x" ]; then
    whatever
fi

if [ "x$var" == "x" ]; then
    whatever
fi

if [ ! "x$var" == "x" ]; then
    whatever
fi

if [ "x$var" != "x" ]; then
    whatever
fi

if [ "X$var" != "X" ]; then
    whatever
fi

if [ "X$var" = "X" ]; then
    whatever
fi

if [ "X$var" == "X" ]; then
    whatever
fi

```

#### Here's what shellcheck currently says:
```
In whatever.sh line 4:
if [ "x$var" = "x" ]; then
     ^-----^ SC2268 (style): Avoid x-prefix in comparisons as it no longer serves a purpose.

Did you mean:
if [ "$var" = "" ]; then


In whatever.sh line 8:
if [ "x$var" == "x" ]; then
     ^-----^ SC2268 (style): Avoid x-prefix in comparisons as it no longer serves a purpose.

Did you mean:
[...]

PR #3249, opened 2025-07-17
x-prefix: add support for "!=" operation and prefixing with X (capital x)

slycordinator:
Changes checkComparisonWithLeadingX to:

1. Work with the "!=" operator in addition to "=" and "==".

2. Support prefixing with "x" and "X".
This is helpful since some scripts have comparisons like [ "X$var" = "X" ] and the like.

Fixes: #2689


koalaman:
This caught quite a few previously unreported issues. Thanks!

slycordinator:
> This caught quite a few previously unreported issues. Thanks!

I just caught how we could simplify it further and possibly catch more unreported instances.

Instead of:
T_Literal id (c:_):_ | toLower c == 'x' ->
and
T_SingleQuoted id (c:rest):_ | toLower c == 'x' ->

We could do:
T_Literal id (c:_):_ | isLetter c ->
and
T_SingleQuoted id (c:rest):_ | isLetter c ->


Since, conceivably someone could do x-prefixing with any letter (and isLetter is likely ever so slightly faster than running toLower before checking which letter it is)

PR #3251, opened 2025-07-18
allow QuickCheck-2.16

juhp:
see also commercialhaskell/stackage#7787


koalaman:
Thanks! I'll test it as part of general release testing.

PR #3264, opened 2025-07-25
Suggest removing null checks in parameter expansions

rhendric:
${parameter-} expands to the empty string if parameter is unset. ${parameter:-} is a more verbose way of saying the same thing, because it expands to the empty string if parameter is unset or empty, and if parameter is empty, it merely replaces one empty with another.

Similarly, ${parameter=} assigns the empty string to parameter if parameter is unset, which is equivalent to what ${parameter:=} does, since assigning the empty string to a parameter already holding the empty string is a no-op.

(Of course, when the replacement to the right of the - or = is non-empty, there is a meaningful difference between the versions with and without the null-check colon.)


koalaman:
Technically correct is obviously the best kind of correct, but pragmatically I suspect the noise of this suggestion will be greater than the impact of fixing the issue, even with the very neat auto-fix functionality. Thanks for the suggestion!

PR #3305, opened 2025-09-19
feat(SC3040): support `set -o pipefail` as specified by POSIX.1-2024

### LINKED ISSUE #2555: `set -o pipefail` is part of POSIX now, consider disabling the warning for `sh`
The POSIX:2022 specification adds support for `set -o pipefail`, so it might make sense to disable this warning even for `#/bin/sh`. 

Source on adding pipefail: POSIX discussion <https://www.austingroupbugs.net/view.php?id=789>, list of shell support <https://unix.stackexchange.com/questions/654885/who-is-responsible-for-providing-set-o-pipefail/654932#654932>, work on `dash` adding it <https://lore.kernel.org/dash/20220422211013.18195-1-chris@chrisn.me.uk/>

As mentioned in stackoverflow, most `sh` implementations support pipefail already, `dash` being about the last big one to not support it. And they are working on it, as mentioned above.

#### For bugs
- Rule Id (if any, e.g. SC1000):  SC3040 (pipefail)
- My shellcheck version (`shellcheck --version` or "online"):  v0.8.0
- [x] The rule's wiki page does not already cover this

```sh
#!/bin/sh
set -o pipefail
```

```
set -o pipefail
       ^-- [SC3040](https://www.shellcheck.net/wiki/SC3040) (warning): In POSIX sh, set option pipefail is undefined.
```

No error

e-kwsm:
fix #2555

### DIFF
--- a/src/ShellCheck/AnalyzerLib.hs
+++ b/src/ShellCheck/AnalyzerLib.hs
@@ -224,9 +224,9 @@ makeParameters spec = params
         hasPipefail =
             case shellType params of
                 Bash -> isOptionSet "pipefail" root
-                Dash -> True
+                Dash -> isOptionSet "pipefail" root
                 BusyboxSh -> isOptionSet "pipefail" root
-                Sh   -> True
+                Sh -> isOptionSet "pipefail" root
                 Ksh  -> isOptionSet "pipefail" root,
         hasExecfail =
             case shellType params of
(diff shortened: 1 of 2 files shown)


eyenx:
Hey @koalaman I was wondering when this gets into a release :-) Thanks for the help & work

PR #3312, opened 2025-09-30
Use https:// for source-repository

cjwatson:
With Cabal >= 3.14, `cabal test` warns:

```
Warning: [git-protocol] Cloning over git:// might lead to an arbitrary code
execution vulnerability. Furthermore, popular forges like GitHub do not
support it. Use https:// or ssh:// instead.
```

https:// should work fine, so use that as suggested.


koalaman:
Thanks!

PR #3318, opened 2025-10-08
build: bump to Haskell2010

e-kwsm:


koalaman:
I've tested it on multiple distros and found no issues. Thanks!

PR #3386, opened 2026-01-03
AI: Allow a function to be referenced in PS4

### LINKED ISSUE #3385: false postive SC2329
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2329
- My shellcheck version (`shellcheck --version` or 'online'): online

- [ ] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

```sh
#! /usr/bin/env bash
#
XTRACE=yep

main() {
    echo "hello:" "$@"
    return 1
}

if [[ "${XTRACE:-}" ]]
then
    PS4func() {
        printf "%b" "+${BASH_LINENO[0]}: "
          
    }
    PS4='\r$(PS4func)'
    set -o xtrace
fi

main "$@"
exit $?
```

```
Line 12	SC2329: This function is never invoked. Check usage (or ignored if invoked indirectly).
```

That function is evoked and the script runs fine.

The exit $? on the last line seems to confuse SC2329.

dotysan:
Closes #3385.

Review required.


koalaman:
This code is uncomfortably tightly coupled to the specific case of assigning a single quoted string regex matching `` `f` `` or `$(f)` to PS4. It does fix the specific case in the bug, but would fail for even minor variations like `PS4="\$(PS4func)"` or `PS4='$(date | PS4func)'`, so it makes it harder to predict behavior.

Unfortunately there isn't really a great approach to handle cases like this at this time. In this case it would be better to suppress with a comment.

PR #3396, opened 2026-01-27
docs: update shellcheck precommit hook version to v0.11.0

mattcristal:
## What was done

Just using the current version from here: https://github.com/koalaman/shellcheck-precommit

I was going to perform a copy-past from the main README and then went to check for the existing pre-commit version.

Updating this to prevent such mistakes.


koalaman:
Thanks!

PR #3410, opened 2026-02-20
Add --files-from option to read input files from a file

kyllianaucher:
Add support for a `--files-from` option to read input file paths from a file.

This allows users to specify a file containing a list of files to analyze,
with support for `-` to read from stdin.

Implementation details:
- Additional file paths are collected before verification.
- Errors while reading are reported via `printErr`.
- `RuntimeException` is thrown if the file cannot be read.
- The existing parsing and analysis pipeline remains unchanged.

The man page has been updated accordingly.

Tested locally with:
- valid file list
- non-existent file
- stdin input (`--files-from=-`)
- combination with direct file arguments

Let me know if you’d like tests or changelog updates.

### DIFF
--- a/shellcheck.1.md
+++ b/shellcheck.1.md
@@ -6,7 +6,7 @@ shellcheck - Shell script analysis tool
 
 # SYNOPSIS
 
-**shellcheck** [*OPTIONS*...] *FILES*...
+**shellcheck** [*OPTIONS*...] [--files-from=FILE] *FILES*...
 
 # DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


bbailleux:
As it is sometimes hard to avoid analysing non script files (e.g. ELF binaries), this feature would be very welcome to avoid false positives in CI/CD

koalaman:
Thanks for the feature and your patience! I further refined it to not give an error with `--files-from=/dev/null` or an otherwise empty list.

PR #3426, opened 2026-03-24
SC2164: suppress when `cd` is the final command in a function 

### LINKED ISSUE #2474: SC2164 false positive with cd at end of shell function
#### For bugs
- Rule Id (if any, e.g. SC1000): SC2164
- My shellcheck version (`shellcheck --version` or "online"): 0.7.0 (installed via apt)
- [X] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2086)
- [X] I tried on https://www.shellcheck.net/ and verified that this is still a problem on the latest commit

```sh
#!/bin/bash
..() {
        cd "$(for ((c=0; c<${1:-1}; ++c)); do echo -n ../; done)"
}
```

```
Line 4:
        cd "$(for ((c=0; c<${1:-1}; ++c)); do echo -n ../; done)"
        ^-- SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
```

I expected no errors to happen because this can not cause an issue (the cd being at the end of the function means it can't continue on to anything else and a return there would be redundant).

notpeter:
Fixes: https://github.com/koalaman/shellcheck/issues/2474

When `cd` is the last command in a function, 
```
#!/bin/bash

foo() {
  cd /abc
}
```

Shellcheck currently triggers [SC2164](https://www.shellcheck.net/wiki/SC2164) and suggest using `cd /abc || return` even though that is no safer than the existing code.

AI disclosure: I used Codex to assist in implementing this PR.

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -3294,6 +3294,7 @@ prop_checkUncheckedPopd10 = verifyNotTree checkUncheckedCdPushdPopd "cd ../.."
 prop_checkUncheckedPopd11 = verifyNotTree checkUncheckedCdPushdPopd "cd ../.././.."
 prop_checkUncheckedPopd12 = verifyNotTree checkUncheckedCdPushdPopd "cd /"
 prop_checkUncheckedPopd13 = verifyTree checkUncheckedCdPushdPopd "cd ../../.../.."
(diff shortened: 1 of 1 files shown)


koalaman:
Definitely a valid complaint, but the added test does not pass

notpeter:
Adjusted so `prop_checkUncheckedCdInFunction1` test case passes and this no longer warns:

```sh
#!/bin/bash
foo() {
  cd /abc
}
```

koalaman:
Thanks!

PR #3439, opened 2026-04-04
build: remove unused process dependency

e-kwsm:


koalaman:
Oh wow, this was used to generate the `man` page in the Cabal 2 build, and has been unnecessary since 2019. Thanks!

PR #3443, opened 2026-04-07
Allow printf -vVAR without a space

### LINKED ISSUE #2650: Relax SC2182 for bash, when printf -v is used
Specifically for *bash*, `printf` has an extra option, `-v` which you can supply a variable name.  This named variable is populated with the result of the `printf` call.  This means you can write `printf -vvar '%s x %s' "$1" "$2"; echo "$var"`.  

Usually, the first arg to `printf` is a format string, so in POSIX shell, an equivalent might be `var="$(printf '%s x %s' "$1" "$1")"; echo "$var"`.

SC2182 fires for *bash* if the first arg is `-vvar`, because `-vvar` does not contain any format string tokens.  However, it is a valid argument to *bash*'s `printf`.

#### For bugs
- Rule Id (if any, e.g. SC1000): SC2182
- My shellcheck version (`shellcheck --version` or "online"): 0.9 (according to https://github.com/vscode-shellcheck/vscode-shellcheck/releases/tag/v0.29.2)

```sh
#!/bin/bash
printf -vvar '%s x %s' "$1" "$2"
echo "$var"
```

[SC2182](https://www.shellcheck.net/wiki/SC2182) (error): This printf format string has no variables. Other arguments are ignored.

no error

dotysan:
Consider this a PS (pull suggestion) as I don't speak Haskell.

I asked AI to fix #2650 and it generated a boatload of spaghetti. I tried to distill it down to this.

Closes #2650.

### DIFF
--- a/src/ShellCheck/Checks/Commands.hs
+++ b/src/ShellCheck/Checks/Commands.hs
@@ -678,10 +678,12 @@ prop_checkPrintfVar19 = verifyNot checkPrintfVar "printf '%(%s)T'"
 prop_checkPrintfVar20 = verifyNot checkPrintfVar "printf '%d %(%s)T' 42"
 prop_checkPrintfVar21 = verify checkPrintfVar "printf '%d %(%s)T'"
 prop_checkPrintfVar22 = verify checkPrintfVar "printf '%s\n%s' foo"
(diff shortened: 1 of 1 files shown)


koalaman:
Looks good to me, thanks!

PR #3445, opened 2026-04-10
feat(SC2035): do not issue SC2035 when script has `set -f` or `set -o noglob`

### LINKED ISSUE #3205: `SC2035` even with `set -f`
I have `set -f` at the top of a script, and I still get a message about potential globbing

e-kwsm:
fix #3205


koalaman:
Thanks! I've never actually seen a `set -f` in real life, so I tend to forget that it exists

PR #3451, opened 2026-04-16
feat: new rule SC2338

Zamiell:
This rule is very similar to SC2292 / require-double-brackets: https://www.shellcheck.net/wiki/SC2292
Which is an optional rule that enforces "[[" over "[" in Bash.

This new rule in this PR is SC2337 / require-double-equals.
It enforces "==" over "=". (It's an optional check in the same way that SC2292 is optional.)

Double equals is preferable because the code becomes more easy to read - in programming, a single equals usually means assignment and a double equals usually means equality comparison.

Let me know if anything else is needed.

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -270,6 +270,13 @@ optionalTreeChecks = [
     }, checkRequireDoubleBracket)
 
     ,(newCheckDescription {
+        cdName = "require-double-equals",
+        cdDescription = "Require == and warn about = in Bash tests",
+        cdPositive = "[[ \"$x\" = \"$y\" ]]",
+        cdNegative = "[[ \"$x\" == \"$y\" ]]"
+    }, checkRequireDoubleEquals)
+
+    ,(newCheckDescription {
         cdName = "check-set-e-suppressed",
         cdDescription = "Notify when set -e is suppressed during function invocation",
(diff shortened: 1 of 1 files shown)


Zamiell:
@koalaman Sorry for the ping, but would you be able to take a look at the PR?

e-kwsm:
potential conflict with #3498: be careful before merging.

src/ShellCheck/Analytics.hs
```diff
@@ -4718,6 +4725,22 @@ checkRequireDoubleBracket params =
...
+prop_checkRequireDoubleEquals4 = verifyNotTree checkRequireDoubleEquals "#!/bin/sh\n[ \"$x\" = \"$y\" ]"
+prop_checkRequireDoubleEquals5 = verifyNotTree checkRequireDoubleEquals "#!/bin/ksh\n[[ \"$x\" = \"$y\" ]]"
+checkRequireDoubleEquals params =
+    if shellType params == Bash
+    then nodeChecksToTreeCheck [check] params
+    else const []
+  where
+    check _ t = case t of
+        TC_Binary id _ "=" _ _ ->
+            style id 2337 "Prefer == over = for string comparisons in Bash tests."
```
e-kwsm:
```suggestion
            style id 2338 "Prefer == over = for string comparisons in Bash tests."
```

e-kwsm (changes requested):

PR #3484, opened 2026-06-11
Exclude underscore from subshell assignment warnings

### LINKED ISSUE #3478: SC2030/SC2031 should not be reported for $_
There seem to be already many reports related to SC2030 and SC2031, but this specific case doesn't seem to be covered, so let me report it here. In short, **could you exclude the variable `$_` from the target of SC2030 and SC2031, just like `$_` is excluded from the target of SC2154**? The variable `$_` is a special shell variable that is set by the shell itself on every execution of a simple command. This means that, when `$_` is referenced, it tries to read the last argument of the immediately previous command (but *not* the value of `$_` set by somewhere inside the subshell as SC2031 suspects). Also, `_` may be used as the assignment target when one wants to *discard* values (e.g., in `read -r a _ c`.), so it should not be reported as SC2030 either.

#### For bugs with existing features

- Rule Id (if any, e.g. SC1000): SC2030, SC2031
- My shellcheck version (`shellcheck --version` or "online"): 0.11.0 (the latest version distributed by Arch Linux)
- [x] I tried on https://www.shellcheck.net/ and verified that this is still a problem on the latest commit.
  <details><summary>Results on https://www.shellcheck.net/</summary>

  ```bash
  #!/usr/bin/env bash

  (
    read -r a _ c <<< 'x y z'
    echo "$a, $c"
  )

  echo 'complicated arguments'
  echo "$_"
  ```

  ```console
  $ shellcheck myscript

  [Line 4:](javascript:setPosition(4, 13))
    read -r a _ c <<< 'x y z'
              ^-- [SC2030](https://www.shellcheck.net/wiki/SC2030) (info): Modification of _ is local (to subshell caused by (..) group).

  [Line 9:](javascript:setPosition(9, 7))
  echo "$_"
        ^-- [SC2031](https://www.shellcheck.net/wiki/SC2031) (info): _ was modified in a subshell. That change might be lost.

  $
  ```
  </details>

```sh
#!/usr/bin/env bash

(
  read -r a _ c <<< 'x y z'
  echo "$a, $c"
)

echo 'complicated arguments'
echo "$_"
```

```console
$ shellcheck pr1208-shellcheck.bash

In pr1208-shellcheck.bash line 4:
  read -r a _ c <<< 'x y z'
            ^-- SC2030 (info): Modification of _ is local (to subshell caused by (..) group).


In pr1208-shellcheck.bash line 9:
echo "$_"
      ^-- SC2031 (info): _ was modified in a subshell. That change might be lost.

For more information:
  https://www.shellcheck.net/wiki/SC2030 -- Modification of _ is local (to su...
  https://www.shellcheck.net/wiki/SC2031 -- _ was modified in a subshell. Tha...
$
```

I do not expect any error.

```console
$ shellcheck pr1208-shellcheck.bash
$
```

VrtxOmega:
Fixes #3478.

## Summary

- exclude `_` from the subshell assignment warning flow alongside the existing special-case variables
- add regression tests for `read -r a _ c` in a subshell and for reading `$_` afterward

## Validation

- `cabal test all` in `haskell:9.6.6`
- `cabal build exe:shellcheck` in `haskell:9.6.6`
- `cabal exec shellcheck -- /evidence/repro-underscore.sh` in `haskell:9.6.6`

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -2069,6 +2069,8 @@ prop_subshellAssignmentCheck20 = verifyTree subshellAssignmentCheck "@test 'foo'
 prop_subshellAssignmentCheck21 = verifyNotTree subshellAssignmentCheck "test1() { echo foo | if [[ $var ]]; then echo $var; fi; }; test2() { echo $var; }"
 prop_subshellAssignmentCheck22 = verifyNotTree subshellAssignmentCheck "( [[ -n $foo || -z $bar ]] ); echo $foo $bar"
 prop_subshellAssignmentCheck23 = verifyNotTree subshellAssignmentCheck "( export foo ); echo $foo"
+prop_subshellAssignmentCheck24 = verifyNotTree subshellAssignmentCheck "( read -r a _ c <<< 'x y z'; ); echo $_"
+prop_subshellAssignmentCheck25 = verifyNotTree subshellAssignmentCheck "( _=discard; ); echo $_"
 subshellAssignmentCheck params t =
     let flow = variableFlow params
         check = findSubshelled flow [("oops",[])] Map.empty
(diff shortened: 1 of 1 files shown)


koalaman:
Thanks!

VrtxOmega:
Thanks again for merging this.

Small process note from our side: Premature Contribution Firewall (PCF) helped us choose this because it was a concrete false positive with no overlapping PR and could stay inside the existing rule/test surface. We'll keep using that gate so future fixes stay small and reviewable.

PR #3488, opened 2026-06-16
Suppress SC2319/SC2320 when $? is assigned to a variable

### LINKED ISSUE #3486: SC2319: rule triggered even though using variable assignment
The resolution text says to assign it to a variable, but I still get the warning when doing so via a guarded statement.

I assume it is expecting variable assignment to look something like:

```sh
[[ -e "./lockfile" ]]
retval=$?
```
However, our scripts run with `-e` set, so the guarded form is needed to avoid script termination.

#### For bugs with existing features

- Rule Id (if any, e.g. SC1000): SC2319
- My shellcheck version (`shellcheck --version` or "online"): 0.11.0, online

```sh
#!/usr/bin/env bash

set -eu -o pipefail

test_function() {
  local retval=0

  [[ -e "./lockfile" ]] || retval=$?

  return "$retval"
}

main() {
  test_function || echo "lockfile does not exist"
}

main "$@"
```

```
In test.sh line 8:
  [[ -e "./lockfile" ]] || retval=$?
                                  ^-- SC2319 (warning): This $? refers to a condition, not a command. Assign to a variable to avoid it being overwritten.

For more information:
  https://www.shellcheck.net/wiki/SC2319 -- This $? refers to a condition, no...
```

No warning

ayushi-work:
## Summary

Fixes #3486

SC2319 warns "This `$?` refers to a condition, not a command. Assign to a variable to avoid it being overwritten." However, it fired even when `$?` was already being assigned to a variable via `||` or `&&`, like:

```sh
[[ -e "./lockfile" ]] || retval=$?
```

The user is already following the advice (assigning to a variable), so the warning is a false positive.

This change skips both SC2319 and SC2320 when `$?` appears inside a `T_Assignment` node.

## Test plan

- [x] `cabal test` passes (all existing + 3 new `prop_` tests)
- [x] The issue snippet no longer triggers SC2319
- [x] Non-assignment cases like `exit $?` after a condition still trigger SC2319

### DIFF
--- a/src/ShellCheck/Analytics.hs
+++ b/src/ShellCheck/Analytics.hs
@@ -5094,9 +5094,14 @@ prop_checkOverwrittenExitCode2 = verifyNot checkOverwrittenExitCode "x; [ $? -eq
 prop_checkOverwrittenExitCode3 = verify checkOverwrittenExitCode "x; echo \"Exit is $?\"; [ $? -eq 0 ]"
 prop_checkOverwrittenExitCode4 = verifyNot checkOverwrittenExitCode "x; [ $? -eq 0 ] && echo Success"
 prop_checkOverwrittenExitCode5 = verify checkOverwrittenExitCode "x; if [ $? -eq 0 ]; then var=$?; fi"
+prop_checkOverwrittenExitCode5a = verify checkOverwrittenExitCode "x; if [ $? -eq 0 ]; then exit $?; fi"
 prop_checkOverwrittenExitCode6 = verify checkOverwrittenExitCode "x; [ $? -gt 0 ] && fail=$?"
+prop_checkOverwrittenExitCode6a = verify checkOverwrittenExitCode "x; [ $? -eq 0 ] && exit $?"
 prop_checkOverwrittenExitCode7 = verifyNot checkOverwrittenExitCode "[ 1 -eq 2 ]; status=$?"
 prop_checkOverwrittenExitCode8 = verifyNot checkOverwrittenExitCode "[ 1 -eq 2 ]; exit $?"
+prop_checkOverwrittenExitCode9  = verifyNot checkOverwrittenExitCode "[[ -e \"./lockfile\" ]] || retval=$?"
+prop_checkOverwrittenExitCode10 = verifyNot checkOverwrittenExitCode "[ -e \"./lockfile\" ] || retval=$?"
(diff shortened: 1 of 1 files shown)


ayushi-work:
Hi @koalaman, just bumping this up!

src/ShellCheck/Analytics.hs
```diff
@@ -5093,10 +5093,13 @@ prop_checkOverwrittenExitCode1 = verify checkOverwrittenExitCode "x; [ $? -eq 1
 prop_checkOverwrittenExitCode2 = verifyNot checkOverwrittenExitCode "x; [ $? -eq 1 ]"
 prop_checkOverwrittenExitCode3 = verify checkOverwrittenExitCode "x; echo \"Exit is $?\"; [ $? -eq 0 ]"
 prop_checkOverwrittenExitCode4 = verifyNot checkOverwrittenExitCode "x; [ $? -eq 0 ] && echo Success"
-prop_checkOverwrittenExitCode5 = verify checkOverwrittenExitCode "x; if [ $? -eq 0 ]; then var=$?; fi"
-prop_checkOverwrittenExitCode6 = verify checkOverwrittenExitCode "x; [ $? -gt 0 ] && fail=$?"
+prop_checkOverwrittenExitCode5 = verify checkOverwrittenExitCode "x; if [ $? -eq 0 ]; then exit $?; fi"
+prop_checkOverwrittenExitCode6 = verify checkOverwrittenExitCode "x; [ $? -eq 0 ] && exit $?"
```
e-kwsm: Please keep the old tests.

```suggestion
prop_checkOverwrittenExitCode5 = verify checkOverwrittenExitCode "x; if [ $? -eq 0 ]; then var=$?; fi"
prop_checkOverwrittenExitCode5a = verify checkOverwrittenExitCode "x; if [ $? -eq 0 ]; then exit $?; fi"
prop_checkOverwrittenExitCode6 = verify checkOverwrittenExitCode "x; [ $? -gt 0 ] && fail=$?"
prop_checkOverwrittenExitCode6a = verify checkOverwrittenExitCode "x; [ $? -eq 0 ] && exit $?"
```

e-kwsm (changes requested):

e-kwsm (approved):

ayushi-work:
Hi! Is there something else that needs to be done?

ayushi-work:
Hi @e-kwsm @koalaman just bumping this up!

ayedalkhshan3-cell (approved):

ayushi-work:
Hi @e-kwsm @koalaman just bumping this up!

PR #3498, opened 2026-07-21
Add SC2337: warn about grep early exit in pipelines under pipefail

### LINKED ISSUE #665: `pipefail` and `grep -q` wont get along well
My colleague @james-masson found an interesting problem when `set -o pipefail` is combined with `grep -q`  and the input to grep is produced in small chunks (like reasonably short line at a time). This combination leads to race condition when process producing input to `grep` may or may not receive SIGPIPE and therefore may or may not produce non 0 exit code.

``` bash
#!/bin/bash

set -e
set -o pipefail

while true
do
ip route  | grep -q "." 
echo "done"
done
```

it would be awesome if shellcheck  warned about it, because in general case combination of `pipefail` and `grep -q` can not produce consistent results

mclasmeier:
**Adds a new linter rule about using `grep -q` in combination with `set -o pipefail`**

When `set -o pipefail` is active and `grep -q` (or `--quiet`) appears as a non-first command in a pipeline, the upstream command may receive SIGPIPE because grep -q exits immediately upon finding a match. This causes the pipeline to report a non-zero exit status under pipefail.

The check covers grep, egrep, and fgrep variants. It does not warn when grep -q reads from a file directly or is the first command in the pipeline.

Please note, my Haskell skills have become a bit rusty -- did this together with Claude.
Is there anything missing for this contribution?

### DIFF
--- a/src/ShellCheck/Checks/Commands.hs
+++ b/src/ShellCheck/Checks/Commands.hs
@@ -104,6 +104,9 @@ commandChecks = [
     ,checkXargsDashi
     ,checkUnquotedEchoSpaces
     ,checkEvalArray
+    ,checkGrepQPipefail
+    ,checkEgrepQPipefail
+    ,checkFgrepQPipefail
     ]
     ++ map checkArgComparison ("alias" : declaringCommands)
     ++ map checkMaskedReturns declaringCommands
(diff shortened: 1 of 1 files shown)


e-kwsm (commented):
This gives false positive to
```sh
grep -e -q  # search "-q"
```

Also, there are other cases where SIGPIPE is sent:

- `grep -m 𝑛 -e 𝑝𝑎𝑡`
- `head`
- sed’s quit e.g. `sed -n -e '/𝑝𝑎𝑡/{p;q;}'`
- nawk’s exit e.g. `nawk '$0 ~ /𝑝𝑎𝑡/ { print; exit; }'`

mclasmeier:
@e-kwsm Hi, thanks for the review.
Regarding the other cases (`head`, `sed`, etc.), I would consider this out of scope for this PR?
Are you generally positive about this contribution, once false positives are addressed?

mclasmeier:
@e-kwsm I have pushed a commit addressing the false positives by using `getOpts` instead.

e-kwsm (commented):
The following non-POSIX options are not covered:

- `grep -m 𝑛 -e pat` (GNU and BSD)
- `grep -L -e pat` (GNU sends SIGPIPE, while BSD does not)

e-kwsm:
> Regarding the other cases (head, sed, etc.), I would consider this out of scope for this PR?

So SC2337 covers grep only, and SC2338? should deal with head?

mclasmeier:
@e-kwsm Updated to also detect `grep -m n` and `grep -L`.
Indeed, I would leave other tools such as `head`, for other PRs.

src/ShellCheck/Checks/Commands.hs
```diff
@@ -400,6 +403,62 @@ checkGrepRe = CommandCheck (Basename "grep") check where
...
+    earlyExitFlags = ["q", "quiet", "m", "max-count", "L"]
+    isEarlyExitFlag name = name `elem` earlyExitFlags
+    parseGrepOpts = getOpts (True, True)
+        "cilLnoqsvwxhHrRbaEFGPe:f:m:A:B:C:d:D:"
+        (map (\name -> (name, True)) longOptionsConsumingParameter)
+    longOptionsConsumingParameter =
+        ["regexp", "file", "max-count", "after-context", "before-context",
+            "context", "directories", "devices"]
+    warnMsg =
+      "In pipefail mode, grep -q may cause the pipeline to fail. Use a non-pipe input like '< <(cmd)', '<<<' or 'grep pattern > /dev/null' instead."
```
e-kwsm: Now flags other than `-q` is checked.

e-kwsm:
This will fix #665.

src/ShellCheck/Checks/Commands.hs
mclasmeier: addressed

src/ShellCheck/Checks/Commands.hs
```diff
@@ -400,6 +403,65 @@ checkGrepRe = CommandCheck (Basename "grep") check where
...
+
+prop_checkEgrepQPipefail1 = verify checkEgrepQPipefail "set -o pipefail; cat file | egrep -q pattern"
+checkEgrepQPipefail = CommandCheck (Basename "egrep") checkQuietGrepInPipefailImpl
+
+prop_checkFgrepQPipefail1 = verify checkFgrepQPipefail "set -o pipefail; cat file | fgrep -q pattern"
+checkFgrepQPipefail = CommandCheck (Basename "fgrep") checkQuietGrepInPipefailImpl
+
+-- Catches occurrences of "grep -q" and variants inside of pipes under pipefail.
+checkQuietGrepInPipefailImpl cmd = do
+    pipefail <- asks hasPipefail
```
e-kwsm: Sorry for nitpicking again.
These no longer target `grep -q`, so something like
`checkQuietGrepInPipefailImpl` → `checkGrepSendsPipefailImpl`?

mclasmeier: I have renamed.

e-kwsm (approved):

e-kwsm:
potential conflict with #3451: be careful before merging.

mclasmeier:
Is there anything else needed from my side?

mclasmeier:
Any update on this? @e-kwsm

e-kwsm:
@koalaman @brother Is it OK to merge?

e-kwsm:
@mclasmeier Thank you for your contribution.

Issue #1844, opened 2020-02-22
Request: Support preprocessor/postprocessor plugins


gromgit:
### For new checks and feature suggestions
- [x] shellcheck.net (i.e. the latest commit) currently gives no useful warnings about this
- [x] I searched through https://github.com/koalaman/shellcheck/issues and didn't find anything related

There are instances where it would be useful to have a user-defined preprocessor flesh out script fragments for Shellcheck to lint, then a postprocessor to read and adjust the results before Shellcheck prepares final judgement. User-defined config scripts that are sourced by application scripts within their own environments are a broad class of files that need this functionality to be properly checked.

The workflow I'm thinking of would look like this (new bits in **bold**):

1. Shellcheck opens source file (or stdin)
2. **Shellcheck writes source file to stdin of preprocessor, and reads modified source from its stdout**
3. Shellcheck parses and checks modified source
4. **Shellcheck sends JSON results to stdin of postprocessor, and reads modified JSON results from its stdout**
5. Shellcheck prints results in desired user format

#1843 is a typical example. Arch Linux's PKGBUILD files are really shell script fragments that are executed in an environment that's [fairly well-defined](https://wiki.archlinux.org/index.php/PKGBUILD), but not accessible outside the build process. A user wanting to use Shellcheck on PKGBUILD files would write the following:

* A preprocessor script (`pbwrap.sh`) that takes the PKGBUILD source from stdin and wrap it like this: https://github.com/vn971/rua/blob/master/res/shellcheck-wrapper
* A postprocessor script (`pbadj.sh`) that takes the JSON results from stdin and uses `jq` or some similar tool to subtract 7 from all `line` and `endLine` properties, to account for the wrapper's prelude, then drop any entries with line numbers that are negative or greater than the original input length. (For this and other reasons, it would be good for the postprocessor to also get the original source and the preprocessed input, probably as command-line arguments.)

Then `shellcheck --pre pbwrap.sh --post pbadj.sh ... my.PKGBUILD` should Get The Job Done.

#### Why not simply write a wrapper around Shellcheck instead?

A Shellcheck wrapper would have to parse all Shellcheck options, to "short-circuit" the above workflow when it doesn't make sense (e.g. `--help`, `--list-optional`, `--version`), and to "count option arguments" otherwise, to ensure all specified files are processed. This would require revisiting the wrapper each time the Shellcheck version changes, in case options are added/renamed/removed. It's not impossible, but it's highly inelegant.

[...]

Issue #356, opened 2015-04-23
Support for pre-declared "external" globals for sourced scripts to avoid SC2154 warnings.


deryni:
We have a series of scripts that operate on a set of shared variables that are created by the top-level script.

The secondary scripts use these variables without declaring them as they are known to exist when run.

shellcheck, obviously, doesn't know this. It would be nice if we could tell it about acceptable external global variables to disable SC2154 warnings about them.

I realize we could use `# shellcheck disable=SC2154` for them but that's a larger hammer than I would like as it will also disable other warnings from typos and the like if they happen to appear on that line.

The `context=sourced` suggestion from https://github.com/koalaman/shellcheck/issues/281 is somewhat related to this request.

koalaman:
Global variables are conventionally in all uppercase (`PATH, DISPLAY, SSH_CLIENT`, etc). For this reason, ShellCheck generally doesn't warn when using undefined uppercase variables. 

The best solution would probably be for ShellCheck to process sourced files though.

deryni:
That's true for system variables but I'm not sure that's as true for script-internal global variables (but maybe it is though that feels like playing in the system's/shell's playground a little bit).

Processing sourced files would be interesting though, assuming you mean when you pass shellcheck the toplevel file it would "recurse" into the sourced files and consider them in that context.

That wouldn't solve all of my cases (we have some scripts that get sourced and others that get run that all expect the external globals) but it would solve some of them.

Moeriki:
When writing scripts for preset systems this uppercase convention can't always be adhered to.

Case in point. I'm writing scripts for Atlassian Bamboo. There is a global variable `bamboo` you can refer to for certain configurational values.

deryni:
Bamboo uses many variables with the `bamboo_` prefix but not a single variable. (Though using a single associative array variable would certainly be nice.) But yes, same idea.

koalaman:
I ran into the same issue with a library set `FLAGS_verbosity` style variables. 

How about a flag for setting a pattern/regex to ignore? This would let you do something like

    export SHELLCHECK_OPTS="--ignore-variable-pattern=FLAGS_|bamboo"

to not get these warnings anymore. Alternatively as a directive, starting the file with

    # shellcheck ignore-variable-pattern=FLAGS_|bamboo

deryni:
[...]

PR #3513, opened 2026-08-05
Support editorconfig

### LINKED ISSUE #1843: Request: Support PKGBUILD files
Arch Linux's [PKGBUILD](https://wiki.archlinux.org/index.php/PKGBUILD) is a shell script with a few differences:
  * It does not have shebang
  * Most variables are used by `makepkg` but are marked as unused (SC2034) by `shellcheck`
  * Some variables (such as `$srcdir`) are provided by `makepkg` but are marked SC2164 by `shellcheck`

[Example PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wasmer-bin&id=fff485450a3e1b2ebc337d0141a8b1f20551e9af)

### LINKED ISSUE #2118: SC2148: add support for bash completions (magic ex comment)
#### For bugs
- SC2148
- My shellcheck version: 0.7.1
- [x] The rule's wiki page does not already cover this (e.g. https://shellcheck.net/wiki/SC2148)

```sh
# POSIX sh(1) completion                                   -*- shell-script -*-

_sh()
{
    local cur prev words cword
    _init_completion || return

    case $prev in
        -c)
            return
            ;;
        -o | +o)
            COMPREPLY=($(compgen -W 'allexport errexit ignoreeof monitor
                noclobber noglob noexec nolog notify nounset verbose vi
                xtrace' -- "$cur"))
            return
            ;;
    esac

    local opts="-a -b -C -e -f -h -i -m -n -o -u -v -x"
    if [[ $cur == -* ]]; then
        COMPREPLY=($(compgen -W "$opts -c -s" -- "$cur"))
        return
    elif [[ $cur == +* ]]; then
        COMPREPLY=($(compgen -W "${opts//-/+}" -- "$cur"))
        return
    fi

    local args ext=
    _count_args "" "@(-c|[-+]o)"
    ((args == 1)) && ext="sh"
    _filedir $ext
} &&
    complete -F _sh sh

# ex: filetype=sh
```

```
In sh line 1:
# POSIX sh(1) completion                                   -*- shell-script -*-
^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
```

Parse either the ``-*- shell-script -*-`` in the first line or the ``# ex: filetype=sh`` in the last line.

#### Some of my thoughts

I know ``-*- shell-script -*-`` is not telling which shell it is maybe checking if the path to the file contains ``/bash-completion/`` works. Or if the first line also contains the string ``completion`` since I think these completions are always bash. It is weird that the ``# ex`` comment says ``filetype=sh`` and then uses bash features but it is so widely used already so yeah.

Here some stats about my /usr/share/bash-completion/completions directory (arch linux):

```
[chiller@arch /usr/share/bash-completion/completions]$ ls | wc -l
947
[chiller@arch /usr/share/bash-completion/completions]$ grep -r "#\!/bin"
bwrap:#!/bin/bash
pkgfile:#!/bin/bash
ipset:#!/bin/bash
runc:#!/bin/bash
pulseaudio:#!/bin/bash
mkinitcpio:#!/bin/bash
[chiller@arch /usr/share/bash-completion/completions]$ grep -r "# ex: " | wc -l
480
[chiller@arch /usr/share/bash-completion/completions]$ grep -rh "# ex:" | sort | uniq
# ex:et ts=2 sw=2 ft=sh
# ex: filetype=sh
# ex: ts=4 sw=4 et filetype=sh
[chiller@arch /usr/share/bash-completion/completions]$ grep -r "shell-script" | wc -l
511
```

I am not sure how this ``ex:`` standard is called and where it is documented but it seems to be wide spread.

I also saw the extension .bash.completion at [pass-otp](https://github.com/tadfisher/pass-otp/blob/develop/pass-otp.bash.completion) maybe also checking for that is an option.

### LINKED ISSUE #2128: Add support for configuring Shellcheck via .editorconfig

#### Scenario
Let's say that there is a project with the following directory setup:
```
|-> Makefile
|-> bin/main.sh
\-> src/*.sh
```

Where depending on the options given to the `make` invocation, different scripts from the `src/` directory are combined together to generate `bin/main.sh`.

In such a scenario, a variable in `src/foo.sh` may only be referenced in `src/bar.sh`, meaning that [SC2034](https://github.com/koalaman/shellcheck/wiki/SC2034) may be perfectly acceptable in context, but would need to be suppressed, as `shellcheck` would generate this warning.

#### Current configuration options
There are several methods by which this warning may be suppressed:
1. Command line options
2. Ignore directives over every instance, in each file
3. Ignore directives under the shebang, in each file
4. Ignore directive in the `.shellcheckrc` file

##### Option 1: Command line options
This is the simplest option, it solves this issue completely *if shellcheck is invoked manually*, however, in this hypothetical scenario this would not be sufficient. For example, if both `make test` and the `.travis.yml` file call shellcheck, then there are two locations where the command line options are specified, and therefore is a duplication of the code that may be unmaintainable. Furthermore, this option would apply to all files in the project, meaning that unintended violations of ignored diagnostics in other files would be missed.

##### Option 2: Ignore directives over every instance, in each file
This option simplifies the invocations, as they can remain simply `shellcheck <regex>.sh`, and also prevents unintended violations from slipping past `shellcheck`. However, this further exacerbates the issue with option 1, as all variables that are intended to be used by other files would need the directive added above every single line that declares a variable that is not used in that file. This is just as unmaintainable as option 1 for this scenario.

##### Option 3: Ignore directives under the shebang, in each file
This is slightly better than option 2, as it only requires one directive per file. However, a new contributor to the project may be unaware of this pattern, or unaware of how `shellcheck` works or what it is. This hypothetical contributor may then add a `src/baz.sh` without the directive under the shebang, thus breaking automated builds. Or, more consqutially, a well meaning contributor, even a well informed one, may know about `shellcheck` and how to use the directives, may blindly copy the directives from another file without knowing which directive disabled which issue. In this case, a directive that is unique to one file may leak to another, and thus an issue may not be caught by shellcheck in this new file.

##### Option 4: Ignore directive in the `.shellcheckrc` file
This option resolves the maintainability option (assuming it is tracked as part of the project's VCS) and prevents developers from mistakenly missing or including the wrong ignore directives, as per the [wiki](https://github.com/koalaman/shellcheck/wiki/Ignore).
> As of v0.7.0 you can create a file .shellcheckrc in your home directory (or your project's base directory), and add disable directives to it.
> ```sh
>$ cat ~/.shellcheckrc
>disable=SC2059
>```

However, the `.shellcheckrc` does not (currently) provide syntax for specifying files by filename or regex. This is fundamentally the issue in this scenario.

#### Solutions
There are two possible methods of configuring which directives to ignore in which files, in a maintainable and version controlled manner:
1. Updating the `.shellcheckrc` syntax
2. Adding support for `.editorconfig`

##### Updating the `.shellcheckrc` syntax
This option really appeals to me as the natural, even "obvious", solution. Since there is already support for a project wide configuration file, it would make sense to extend this functionality to allow specifying certain directives to specific files. However, updating the syntax may affect existing `.shellcheckrc` files, so the specific changes to the syntax should be backwards compatible.

##### Adding support for `.editorconfig`
For an equivalent hypothetical project in C#, for example, a (mostly) analogous warning to [SC2034](https://github.com/koalaman/shellcheck/wiki/SC2034) would be [IDE0051](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0051), and to disable this warning for specific files would be as simple as adding the directive to the project's `.editorconfig`:
```editorconfig
[src/*.cs]
dotnet_diagnostic.IDE0051.severity = none
```

So for `shellcheck`, this hypothetical project could add something like this to their `.editorconfig`:
```editorconfig
shellcheck_SC2034 = disabled
```

This has three benefits over updating the `.shellcheckrc` syntax.
1. It consolidates all linting configurations into one file; `.editorconfig`, which is [widely recognized](https://editorconfig.org/#download).
2. The syntax is already well established, meaning that the `.shellcheckrc` syntax is free to be updated in the future in a divergent way.
3. `.editorconfig` files use the [INI](https://en.wikipedia.org/wiki/INI_file) format, for which Haskell already has a parsing library, [Data.Ini](https://hackage.haskell.org/package/ini-0.4.1/docs/Data-Ini.html)

Freed-Wu:
Fix #1843, Fix #2128, Fix #2118

> EditorConfig cores shall accept and report all syntactically valid key-value pairs, even if the key is not defined in this specification.
> EditorConfig plugins shall ignore unrecognized keys and invalid/unsupported values.

So we can use `.editorconfig` or global `~/.config/editorconfig.ini` :

```ini
[{build,*.subpackage}.sh]
shellcheck.disable=SC2034

[{*.ebuild,*.eclass,*.conf,color.map,.devscripts,*.mdd}]
shellcheck.shell=bash
shellcheck.disable=SC2034

[{PKGBUILD,*.install}]
shellcheck.shell=bash
shellcheck.disable=SC2034,SC2154

[APKBUILD]
shellcheck.shell=sh
shellcheck.disable=SC2034,SC2154

[*/bash-completion/completions/*]
shellcheck.shell=bash
```

Refer [termux-language-server](https://github.com/termux/termux-language-server) for filenames.

Final result:

Screenshot_20260805_211713

PS: If someone want to realize #1844 #356 , can extend the config:

```ini
[{PKGBUILD,*.install}]
shellcheck.shell=bash
shellcheck.sourcefile=/the/path/of/PKGBUILD.d.sh
```

`PKGBUILD.d.sh`:

```sh
# avoid SC2154
export pkgdir srcdir
# avoid SC2034
export pkgver # ...
```

haskell doesn't have library for editorconfig. and ini doesn't support disabling inline comment as editorconfig. So we have to create one.

### DIFF
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -78,7 +78,8 @@ data Options = Options {
     sourcePaths      :: [FilePath],
     formatterOptions :: FormatterOptions,
     minSeverity      :: Severity,
-    rcfile           :: Maybe FilePath
+    rcfile           :: Maybe FilePath,
+    fileNameOverride :: Maybe FilePath
 }
 
 defaultOptions = Options {
@@ -89,7 +90,8 @@ defaultOptions = Options {
         foColorOption = ColorAuto
     },
     minSeverity = StyleC,
-    rcfile = Nothing
+    rcfile = Nothing,
+    fileNameOverride = Nothing
 }
 
 usageHeader = "Usage: shellcheck [OPTIONS...] FILES..."
@@ -138,7 +140,10 @@ options = [
         (NoArg $ Flag "help" "true") "Show this usage summary and exit",
     Option "" ["files-from"]
         (ReqArg (Flag "files-from") "FILE")
-        "Read input files from FILE (one per line, or '-' for stdin)"
+        "Read input files from FILE (one per line, or '-' for stdin)",
+    Option "" ["file-name"]
+        (ReqArg (Flag "file-name") "FILE")
+        "Use FILE as the filename for parsing EditorConfig configuration when input is stdin"
     ]
 getUsageInfo = usageInfo usageHeader options
 
@@ -421,6 +426,11 @@ parseOption flag options =
                 rcfile = Just str
             }
 
+        Flag "file-name" str -> do
+            return options {
+                fileNameOverride = Just str
+            }
+
         Flag "enable" value ->
             let cs = checkSpec options in return options {
                 checkSpec = cs {
@@ -519,8 +529,12 @@ ioInterface options files = do
     -- merged with any shellcheck.* directives found in applicable
     -- EditorConfig files.
     getConfig cache filename = do
-        rcResult <- getRcConfig cache filename
-        ecResult <- getEditorConfig filename
+        let configFilename =
+                if filename == "-"
+                then fromMaybe filename (fileNameOverride options)
+                else filename
+        rcResult <- getRcConfig cache configFilename
+        ecResult <- getEditorConfig configFilename
         return $ mergeConfigs filename rcResult ecResult
 
     mergeConfigs filename rcResult ecResult =
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -300,10 +300,19 @@ runFormatter sys format options files = do
             }
             result <- checkScript sys checkspec
             onResult format result sys
+            -- A malformed EditorConfig (invalid 'root' or 'shellcheck.*'
+            -- directive) means shellcheck cannot apply the requested
+            -- configuration, so it fails rather than silently proceeding.
             return $
-                if null (crComments result)
-                then NoProblems
-                else SomeProblems
+                if any editorConfigError (crComments result)
+                    then SupportFailure
+                    else if null (crComments result)
+                        then NoProblems
+                        else SomeProblems
+
+        editorConfigError pc =
+            cCode (pcComment pc) == 1134 &&
+                ".editorconfig" `isSuffixOf` posFile (pcStartPos pc)
 
 parseEnum name value list =
     case lookup value list of
@@ -535,6 +544,10 @@ ioInterface options files = do
                 else filename
         rcResult <- getRcConfig cache configFilename
         ecResult <- getEditorConfig configFilename
+        -- A rejected EditorConfig blob (one that only contains invalid
+        -- 'root'/'shellcheck.*' rejections) is surfaced with the
+        -- EditorConfig source so its SC1134 is attributable to it; this
+        -- also lets the formatter treat it as a fatal config error.
         return $ mergeConfigs filename rcResult ecResult
 
     mergeConfigs filename rcResult ecResult =
@@ -542,7 +555,14 @@ ioInterface options files = do
             (Nothing, Nothing) -> Nothing
             (Just (rcPath, rc), Nothing) -> Just (rcPath, rc)
             (Nothing, Just (ecPath, ec)) -> Just (ecPath, ec)
-            (Just (rcPath, rc), Just (_, ec)) -> Just (rcPath, rc ++ "\n" ++ ec)
+            (Just (rcPath, rc), Just (ecPath, ec)) ->
+                if isEditorConfigRejection ec
+                    then Just (ecPath, ec)
+                    else Just (rcPath, rc ++ "\n" ++ ec)
+
+    isEditorConfigRejection ec =
+        all (\l -> null (trim l) || l == "invalid editorconfig value") (lines ec)
+
 
     getRcConfig cache filename =
         case rcfile options of
@@ -589,17 +609,20 @@ ioInterface options files = do
             [] -> Nothing
             ((sourceFile, _):_) -> Just (sourceFile, concatMap snd contributions)
       where
-        -- For each EditorConfig file: either report an invalid 'root'
-        -- declaration at its line, or yield the matching shellcheck.*
-        -- directives. Invalid roots take priority over directives.
+        -- For each EditorConfig file: report any invalid 'root'
+        -- declaration (which takes priority) as a rejected config blob,
+        -- otherwise yield the matching shellcheck.* directives
+        -- (invalid directives are reported by editorConfigDirectives as a
+        -- rejected blob so the .shellcheckrc parser emits SC1134).
         directivesFor path (file, contents) =
-            case invalidRootLines contents of
+            let relative = makeRelativeTo (takeDirectory file) path
+            in case invalidRootLines contents of
                 (badLine:_) ->
                     [(file, replicate (badLine - 1) '\n' ++ "invalid editorconfig value\n")]
                 [] ->
-                    let relative = makeRelativeTo (takeDirectory file) path
-                        result = parseEditorConfig contents relative
-                    in if null result then [] else [(file, result)]
+                    case editorConfigDirectives contents relative of
+                        Nothing -> []
+                        Just blob -> [(file, blob)]
 
         makeRelativeTo dir path =
             case stripPrefix (addTrailingSlash dir) path of
--- a/shellcheck.hs
+++ b/shellcheck.hs
@@ -20,6 +20,7 @@
 import qualified ShellCheck.Analyzer
 import           ShellCheck.Checker
 import           ShellCheck.Data
+import           ShellCheck.EditorConfig
 import           ShellCheck.Interface
 import           ShellCheck.Regex
 
@@ -110,7 +111,7 @@ options = [
     Option "" ["list-optional"]
         (NoArg $ Flag "list-optional" "true") "List checks disabled by default",
     Option "" ["norc"]
-        (NoArg $ Flag "norc" "true") "Don't look for .shellcheckrc files",
+        (NoArg $ Flag "norc" "true") "Don't look for .shellcheckrc and .editorconfig files",
     Option "" ["rcfile"]
         (ReqArg (Flag "rcfile") "RCFILE")
         "Prefer the specified configuration file over searching for one",
@@ -514,8 +515,22 @@ ioInterface options files = do
         fallback path _ = return path
 
 
-    -- Returns the name and contents of .shellcheckrc for the given file
-    getConfig cache filename =
+    -- Returns the name and contents of .shellcheckrc for the given file,
+    -- merged with any shellcheck.* directives found in applicable
+    -- EditorConfig files.
+    getConfig cache filename = do
+        rcResult <- getRcConfig cache filename
+        ecResult <- getEditorConfig filename
+        return $ mergeConfigs filename rcResult ecResult
+
+    mergeConfigs filename rcResult ecResult =
+        case (rcResult, ecResult) of
+            (Nothing, Nothing) -> Nothing
+            (Just (rcPath, rc), Nothing) -> Just (rcPath, rc)
+            (Nothing, Just (ecPath, ec)) -> Just (ecPath, ec)
+            (Just (rcPath, rc), Just (_, ec)) -> Just (rcPath, rc ++ "\n" ++ ec)
+
+    getRcConfig cache filename =
         case rcfile options of
             Just file -> do
                 -- We have a specified rcfile. Ignore normal rcfile resolution.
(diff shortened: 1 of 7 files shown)


e-kwsm (changes requested):
This does not search .editorconfig&rsquo;s in parental directories when `root=false`.

Also, this is not compliant with <https://editorconfig.org/#file-location>

> EditorConfig files are read top to bottom and the most recent rules found take precedence.

```console
$ cat > .editorconfig << 'EOF'
root = true

[*]
shellcheck.shell=sh

[foo]
shellcheck.shell=bash
EOF
$ echo 'pushd foo || exit' > foo
$ shellcheck foo

In foo line 1:
pushd foo || exit
^-------^ SC3044 (warning): In POSIX sh, 'pushd' is undefined.

For more information:
  https://www.shellcheck.net/wiki/SC3044 -- In POSIX sh, 'pushd' is undefined.
```

shellcheck.hs
e-kwsm:
```suggestion
        (NoArg $ Flag "norc" "true") "Don't look for .shellcheckrc and .editorconfig files",
```

e-kwsm (changes requested):

e-kwsm (changes requested):
Without .editorconfig, shellcheck issues SC2148:

```console
$ echo 'pushd foo || exit' > x
$ shellcheck x

In x line 1:
pushd foo || exit
^-- SC2148 (error): Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.

For more information:
  https://www.shellcheck.net/wiki/SC2148 -- Tips depend on target shell and y...
```

But when .editorconfig specifies a *wrong* shell, SC2148 is _not_ issued

```console
$ > .editorconfig cat << 'EOF'
root = true
[*]
shellcheck.shell = zsh
# XXX: zsh is not supported
EOF
$ shellcheck     x  # no warning
$ shellcheck - < x  # stdin, neither
```

e-kwsm (changes requested):
```ini
root =
```

→ No warnings/errors are issued.

```ini
root = true
[*]
shellcheck.shell =
```

→ SC1134 is issued but the location is incorrect
```console
$ shellcheck -s bash x

In x line 1:
pushd foo || exit
^-- SC1134 (error): Failed to process x, line 1:  Fix any mentioned problems and try again.

For more information:
  https://www.shellcheck.net/wiki/SC1134 -- Failed to process x, line 1:  Fix...
```

```ini
root = true
[*]
shellcheck.disable = abc
```

→ the same as above

src/ShellCheck/EditorConfig.hs
```diff
@@ -0,0 +1,195 @@
...
+        in case rest' of
+            (']':rest'') -> "[" ++ translateClass cls ++ "]" ++ go rest''
+            _ -> "\\[" ++ go rest
+    go ('{':rest) =
+        let (body, rest') = break (== '}') rest
+        in case rest' of
+            ('}':rest'') ->
+                "(" ++ intercalate "|" (map go (splitCommas body)) ++ ")" ++ go rest''
+            _ -> "\\{" ++ go rest
+    go (c:rest)
```
e-kwsm, line 268: This does not support [`{num1..num2}` pattern](https://spec.editorconfig.org/#glob-expressions).

e-kwsm (changes requested):
When a file is a symlink, EditorConfig rules are applied based on its *realpath*:

```console
$ cat > .editorconfig << EOF
root = true

[foo]
shellcheck.shell=sh

[bar]
shellcheck.shell=bash
EOF
$ cat > foo << EOF
pushd foo || exit
EOF
$ ln -s foo bar
$ shellcheck -f gcc foo  # expected
foo:1:1: warning: In POSIX sh, 'pushd' is undefined. [SC3044]
$ shellcheck -f gcc bar  # ?
bar:1:1: warning: In POSIX sh, 'pushd' is undefined. [SC3044]
$ rm bar; cp foo bar
$ shellcheck -f gcc bar  # expected
```

src/ShellCheck/EditorConfig.hs
```diff
@@ -0,0 +1,195 @@
...
+        in case break (== '=') t of
+            (key, '=':value) ->
+                let key' = trim key
+                    value' = trim value
+                in if "shellcheck." `isPrefixOf` key'
+                    then Just (drop (length "shellcheck.") key', value')
+                    else Nothing
+            _ -> Nothing
+
+    stripComment = takeWhile (\c -> c /= '#' && c /= ';')
```
e-kwsm: [Inline comments are not allowed](https://spec.editorconfig.org/#no-inline-comments).

e-kwsm (changes requested):

src/ShellCheck/EditorConfig.hs
```diff
@@ -0,0 +1,294 @@
...
+    -- '{1..3}' which match any integer in the range.
+    braceAlternatives body =
+        case break (== '.') body of
+            (fromStr, '.':'.':toStr)
+                | Just from <- readInt fromStr
+                , Just to <- readInt toStr ->
+                    map show $
+                        if from <= to
+                        then [from..to]
+                        else [from, from-1..to]
```
e-kwsm: <https://spec.editorconfig.org/#glob-expressions>:

> `num1` is required to be less than `num2`. For instance, `{1..3}`, `{-1..4}` are valid, but `{-4..-5}`, `{3..1}` are not.

e-kwsm (changes requested):

src/ShellCheck/EditorConfig.hs
```diff
@@ -0,0 +1,294 @@
...
+    -- Expand EditorConfig brace alternatives. Besides the comma
+    -- separated form, EditorConfig supports numeric ranges like
+    -- '{1..3}' which match any integer in the range.
+    braceAlternatives body =
+        case break (== '.') body of
+            (fromStr, '.':'.':toStr)
+                | Just from <- readInt fromStr
+                , Just to <- readInt toStr ->
+                    map show $
+                        if from <= to
```
e-kwsm: <https://spec.editorconfig.org/#glob-expressions>:

> any integer numbers between `num1` and `num2`, where `num1` and `num2` can be either positive or negative. `num1` is required to be less than `num2`.

```suggestion
                        if from < to
```

src/ShellCheck/EditorConfig.hs
```diff
@@ -0,0 +1,294 @@
...
+                        then [from..to]
+                        else []
```
e-kwsm, line 313: This silently ignores malformed patterns.

Freed-Wu: `num1 >= num2` is invalid, however, editorconfig standard has no regulations about how to handle invalid behaviour. Silent ignore it is a legal behaviour. Refer https://github.com/editorconfig/editorconfig-core-py/blob/e1298e0b8c66041f2e32e84d7c7bbe99ffdba339/editorconfig/fnmatch.py#L109

e-kwsm (changes requested):
```editorconfig
root = true

[foo{,bar}]
shellcheck.shell = sh
```

https://github.com/editorconfig/editorconfig-core-py v0.17.1 accepts the above .editorconfig, but shellcheck fails:

```console
$ shellcheck foo
shellcheck: Explicit error in module Text.Regex.TDFA.String : Text.Regex.TDFA.String died: parseRegex for Text.Regex.TDFA.String failed:"^(.*/)?foo(|bar)$" (line 1, column 12):
unexpected "|"
expecting empty () or anchor ^ or $ or an atom
CallStack (from HasCallStack):
  error, called at lib/Text/Regex/TDFA/Common.hs:31:3 in regex-tdfa-1.3.2.5-K3WM9QKkG8i6YZ8c6M2glc:Text.Regex.TDFA.Common
```

regex-tdfa does not seem to allow an empty branch, but I&rsquo;m not sure if this is a feature or not.

e-kwsm:
`foo(()|bar)` and `foo(bar)?` work around.

e-kwsm (commented):
does not seem to work with nested pattern:

```editorconfig
root = true
[{foo,ba{r,z}}]
shellcheck.shell = sh
```

Freed-Wu:
> regex-tdfa does not seem to allow an empty branch, but I’m not sure if this is a feature or not.

I doubt it is a bug of regex-tdfa.

e-kwsm (changes requested):
`shellcheck.shell=invalid` (https://github.com/koalaman/shellcheck/pull/3513#pullrequestreview-4882537701) is not addressed yet (51df507973a45adc2b064bf55738a496bb20c5e7).

src/ShellCheck/EditorConfig.hs
```diff
@@ -0,0 +1,340 @@
...
+                    else Nothing
+            _ -> Nothing
+
+    -- Only emit directives that the .shellcheckrc parser can handle.
+    -- Unknown shells (e.g. 'shellcheck.shell=zsh') would otherwise
+    -- silently suppress the SC2148 "unknown shell" warning, so they are
+    -- dropped here. Empty values are kept: the rc parser rejects them,
+    -- producing an SC1134 error at the right file and line.
+    isUsableDirective "shell" value = null value || isJust (shellForExecutable value)
+    isUsableDirective _ value = not (null value)
```
e-kwsm: Hmm, I prefer to emit `Unknown shell: zsh` and exit with four.

e-kwsm: Something like

```
shellcheck.shell = zsh
^-- SC1071 (error): ShellCheck only supports sh/bash/dash/ksh/'busybox sh' scripts. Sorry!
```

e-kwsm (commented):
```editorconfig
root = true

[foo]
shellcheck.disable = #abc
#       ^~~ no error is issued (unexpected)

[bar]
shellcheck.disable = abc
#       ^-- SC1134 (error): Failed to process /tmp/.editorconfig, line 8:  Fix any mentioned problems and try again.
```

e-kwsm (commented):
```editorconfig
root = true

[foo]
shellcheck.disable = SC2148 #abc
#       ^-- no error is issued

[bar]
shellcheck.disable = #abc
#       ^-- SC1134 (error): Failed to process /tmp/.editorconfig, line 8: Expected '=' after directive key. Fix any mentioned problems and try again.
#       Here, its value is nonempty, "#abc"
```

https://github.com/koalaman/shellcheck/blob/9af7ee28ce587baadd950b85dd6826a16b9c068d/src/ShellCheck/Parser.hs#L1024

It seems that a value of `shellcheck.disable` is parsed as is in the EditorConfig module, but it is interpreted in the same way as .shellcheckrc, and eventually trailing part after `#` is trimmed.

Freed-Wu:
> it is interpreted in the same way as .shellcheckrc

So should we provide a different error message?

e-kwsm:
> So should we provide a different error message?

As of 9aa787133fc293470ee05961ed893fbdcbce4533, shellcheck emits

```
In /tmp/.editorconfig line 4:
shellcheck.disable = SC2148 #abc
       ^-- SC1134 (error): Failed to process /tmp/.editorconfig, line 4: Expected '=' after directive key. Fix any mentioned problems and try again.
```

which is clear.

e-kwsm (approved):

e-kwsm:
Thank you for your contribution!

